PR #1, opened 2017-03-15
Add RAML file documenting API.

forevermatt:


api.raml
```diff
@@ -0,0 +1,107 @@
+#%RAML 1.0
+title: ID Broker API
+protocols: [ HTTPS ]
```
fillup: May need to say http and https so in test mode it can just listen on http, especially since the container only listens on http/80 and https will be handled by the load balancer.

forevermatt: Is there a way we can design it to always use https (even behind the load balancer)? Since we're passing around passwords in plaintext, always using https (even on a private network) seems like a good practice, especially if it's easy.

api.raml
```diff
@@ -0,0 +1,107 @@
...
+mediaType: application/json
+securitySchemes:
+  # TODO: Define auth. bearer token schema
```
forevermatt: Exactly

api.raml
```diff
@@ -0,0 +1,107 @@
...
+            example: |
+              {
+                "employee_id": "12345",
+                "first_name": "John",
+                "last_name": "Smith",
+                "display_name": "John Smith",
+                "username": "john_smith",
+                "email": "john_smith@example.com",
+                "active": "yes",
+                "locked": "no"
```
forevermatt: @fillup Do we want these timestamps coming out? It doesn't seem like we would need (want?) them on an authentication call, but I don't know what other situations you might have had in mind.

api.raml
```diff
@@ -0,0 +1,107 @@
...
+              }
+      400:
+        description:
+          The given credentials were NOT acceptable/valid (i.e. - the
+          authentication failed).
+
```
forevermatt: @fillup What kind of interface do we want to specify for this part of the conversation between the called (such as SilAuth) and ID Broker? Do we want/need to divulge whether the user didn't exist vs. a bad password was provided? SilAuth at this point doesn't care, but I don't know what would be the best design just from looking at the ID Broker by itself.

api.raml
```diff
@@ -0,0 +1,107 @@
...
+    body:
+      application/json:
+        type: User
+        example: |
+          {
+            "employee_id": "12345",
+            "first_name": "John",
+            "last_name": "Smith",
+            "username": "john_smith",
+            "email": "john_smith@example.com"
```
forevermatt: Yep, those can be provided as well.

api.raml
```diff
@@ -0,0 +1,107 @@
...
+          }
+    responses:
+      200:
+        description:
+          The corresponding user record (matching on `employee_id`) was either
+          created (if it did not yet) or updated to match the given data (if it
+          did exist).
```
forevermatt: Ah, good catch.

forevermatt: Added.

PR #6, opened 2017-03-21
Feature/initial features

wcjr:
Since there's been so much activity from teammates depending upon this code, I wanted to go ahead and get the core features back on `develop` as soon as possible.  I'll continue on the following efforts in other branches built off this branch:

- Complete API tests
- Complete README
- Complete all remaining TODO's in code and [issues](https://github.com/silinternational/idp-id-broker/issues)
- Review [Yii best practices](http://www.yiiframework.com/doc-2.0/guide-security-best-practices.html)
- Implement logging as needed
- Implement email as needed


Dockerfile
```diff
@@ -27,9 +27,7 @@ COPY application/ /data/
 
 # Fix folder permissions
 RUN chown -R www-data:www-data \
-    console/runtime/ \
-    frontend/runtime/ \
-    frontend/web/assets/
```
forevermatt: I ended up having to create these two folders when I brought this repo up (to fix an error about missing folders), so if you were simply removing these lines because the folders didn't exist, that might not be necessary now.

Makefile
```diff
@@ -1,52 +1,28 @@
...
 
-composerupdate:
-	docker-compose run --rm --user="0:0" cli composer update
+depsupdate:
+	docker-compose run --rm cli composer update
 
-rmDb:
-	docker-compose kill db
-	docker-compose rm -f db
+depsrefresh: depsupdate deps
```
forevermatt: Unless I'm mistaken, running `composer update` both updates and installs the dependencies, so I don't think `depsrefresh` is necessary.

application/common/config/main.php
```diff
@@ -1,123 +1,32 @@
...
-$alertsEmailEnabled = Env::get('ALERTS_EMAIL_ENABLED');
-$fromEmail = Env::get('FROM_EMAIL');
-$appEnv = Env::get('APP_ENV');
-$uiCorsOrigin = Env::get('UI_CORS_ORIGIN');
+$mysqlHost     = getRequiredEnv('MYSQL_HOST');
+$mysqlDatabase = getRequiredEnv('MYSQL_DATABASE');
+$mysqlUser     = getRequiredEnv('MYSQL_USER');
+$mysqlPassword = getRequiredEnv('MYSQL_PASSWORD');
+
+function getRequiredEnv($name)
```
forevermatt: Do you think it would be worth adding a function like this to our `PhpEnv` library?

Dockerfile
wcjr: I'd like to see what's either creating those or requiring them, I don't think they're needed anymore...I'll blow everything away again on my side just to make sure it works properly from scratch.

wcjr: Looks like it was a timing issue off of `develop`, you should not see that problem using this branch.

Makefile
wcjr: thanks, made that correction.

application/common/config/main.php
wcjr: I had the same thought, I'd love to see it there if you think it's a common need.

application/common/helpers/Utils.php
```diff
@@ -0,0 +1,10 @@
+<?php
+namespace common\helpers;
+
+class Utils
```
forevermatt: What would you think of making this a more meaningful class (e.g. `Time`, `UtcTime`, or similar)? I ask because I read (or heard) someone articulate that a Utils class is itself a code smell, and that seems accurate to me.

application/common/models/PasswordHistory.php
```diff
@@ -0,0 +1,56 @@
...
+
+class PasswordHistory extends PasswordHistoryBase
+{
+    public $password;
+
+    public function rules(): array
+    {
+        return ArrayHelper::merge([
+            [
+                'created_utc', 'default', 'value' => Utils::now(),
```
forevermatt: Since `default` only applies if the the existing value is empty, should we instead do something that will force `created_utc` to be the current datetime?

application/common/config/main.php
forevermatt: Seems like a fine feature to have.

application/common/models/User.php
```diff
@@ -0,0 +1,190 @@
...
+
+class User extends UserBase
+{
+    const SCENARIO_SAVE            = 'save';
+    const SCENARIO_UPDATE_PASSWORD = 'update_password';
+    const SCENARIO_AUTHENTICATE    = 'authenticate';
+
+    public $password;
+
+    public function scenarios(): array
```
forevermatt: I couldn't tell by a quick reading of [the docs for `scenarios()`](http://www.yiiframework.com/doc-2.0/yii-base-model.html#scenarios()-detail), but if a given scenario doesn't include a specific field, can it be massively assigned or not? It sounded like it would simply not be validated... which seems odd, and isn't what we want (I think). The comment in the docs about using `!fieldName` to make it non-massively-assignable is what makes me wonder...

application/common/models/User.php
```diff
@@ -0,0 +1,190 @@
...
+                'password', 'required',
+                'on' => [self::SCENARIO_UPDATE_PASSWORD, self::SCENARIO_AUTHENTICATE],
+            ],
+            [
+                'password', 'string',
+            ],
+            [
+                'password',
+                $this->validatePassword(),
+                'on' => self::SCENARIO_AUTHENTICATE,
```
forevermatt: I'm still having some trouble wrapping my mind around exactly what's happening when we use these validation rules and specific scenarios for things like validating a username and password. These validation rules still feel like they were intended for creating/updating database records, and while we technically can use them for validating a given username and password, it seems to make it harder to clearly see exactly what's going on... and validating a login is one of those situations where I keep wanting to be able to see exactly what's happening, as clearly as possible. Thoughts?

application/common/models/User.php
wcjr: if they're not referenced in that scenario they cannot be massively assigned.  I believe the `!attribute` is some special case of it being active without validation or something strange like that.

application/common/models/User.php
```diff
@@ -0,0 +1,190 @@
...
+    {
+        return [
+            'changeTracker' => [
+                'class' => AttributeBehavior::className(),
+                'attributes' => [
+                    ActiveRecord::EVENT_BEFORE_INSERT => 'last_changed_utc',
+                    ActiveRecord::EVENT_BEFORE_UPDATE => 'last_changed_utc',
+                ],
+                'value' => Utils::now(),
+                'skipUpdateOnClean' => true, // only update the column if the model is dirty
```
forevermatt: Did you find a way to use these behaviors such that a change to `last_synced_utc` wouldn't trigger a change to this `last_changed_utc`?

application/common/models/User.php
```diff
@@ -0,0 +1,190 @@
...
+            ],
+            'updateTracker' => [
+                'class' => AttributeBehavior::className(),
+                'attributes' => [
+                    ActiveRecord::EVENT_BEFORE_INSERT => 'last_synced_utc',
+                    ActiveRecord::EVENT_BEFORE_UPDATE => 'last_synced_utc',
+                ],
+                'value' => Utils::now(),
+                'skipUpdateOnClean' => false, // always update the column regardless of dirtiness
```
forevermatt: Are there scenarios where the model will change (such as a new password pushed from the password-reset tool) but where we _don't_ want `last_synced_utc` to change? I think `last_synced_utc` is just for when this record has been synchronized with the ID Store, so changing the password shouldn't update `last_synced_utc` (as far as I understand things).

application/common/models/User.php
```diff
@@ -0,0 +1,190 @@
...
+            'employee_id',
+            'first_name',
+            'last_name',
+            'display_name',
+            'username',
+            'email',
+            'active',
+            'locked',
+            'last_changed_utc',
+            'last_synced_utc',
```
forevermatt: We probably need to check with @fillup about whether to include the changed/synced fields in responses. I left them out of the RAML spec for the API because I still don't know of any use case for sending them, so we'll just need to figure out which way we want to go.

application/common/models/User.php
```diff
@@ -0,0 +1,190 @@
...
+        try {
+            if ($this->hasPasswordAlready()) {
+                if (! $this->saveHistory()) {
+                    return false;
+                }
+            }
+
+            $this->password_hash = password_hash($this->password, PASSWORD_DEFAULT);
+
+            if (! parent::save($runValidation, $attributeNames)) {
```
forevermatt: Is there a way we can tell just from the code here in this `savePassword(...)` function that this call to `parent::save(...)` is only going to update the password-related field(s) of the User (and not any/all of the other fields)?

application/common/models/User.php
```diff
@@ -0,0 +1,190 @@
...
+    {
+        return ! empty($this->password_hash);
+    }
+
+    private function saveHistory(): bool
+    {
+        $history = new PasswordHistory();
+
+        $history->user_id = $this->id;
+        $history->password = $this->password;
```
forevermatt: Ahhh... I get it. You're giving the PasswordHistory object the password (not just he password_hash) to check for reuse. Interesting. I would have just defaulted to doing it in the User class. What led you to doing it here instead?

application/common/models/User.php
```diff
@@ -0,0 +1,190 @@
...
+                'password',
+                $this->validatePassword(),
+                'on' => self::SCENARIO_AUTHENTICATE,
+            ],
+            [
+                ['last_synced_utc', 'last_changed_utc'],
+                'default', 'value' => Utils::now(),
+            ],
+            [
+                ['employee_id', 'first_name', 'last_name', 'display_name', 'username', 'email'],
```
forevermatt: It looks like the parent::rules() already include a rule that these fields (plus `password_hash`) have to be strings. Was there still a need to include that rule here, too?

application/common/models/UserBase.php
```diff
@@ -0,0 +1,79 @@
...
+            'first_name' => Yii::t('app', 'First Name'),
+            'last_name' => Yii::t('app', 'Last Name'),
+            'display_name' => Yii::t('app', 'Display Name'),
+            'username' => Yii::t('app', 'Username'),
+            'email' => Yii::t('app', 'Email'),
+            'password_hash' => Yii::t('app', 'Password Hash'),
+            'active' => Yii::t('app', 'Active'),
+            'locked' => Yii::t('app', 'Locked'),
+            'last_changed_utc' => Yii::t('app', 'Last Changed Utc'),
+            'last_synced_utc' => Yii::t('app', 'Last Synced Utc'),
```
forevermatt: It probably doesn't matter, but I tend to update field labels like these '... Utc' ones in the User class so that they end up as '... (UTC)'.

application/console/config/main.php
```diff
@@ -2,23 +2,8 @@
 return [
     'id' => 'app-console',
     'basePath' => dirname(__DIR__),
-    'bootstrap' => ['log', 'gii'],
-    'controllerNamespace' => 'console\controllers',
```
forevermatt: Do we know if we plan to have any CLI commands for this? Let's see, I think the call to batch-sync with the ID store will be on a cron job, so we probably will. We may need this `controllerNamespace` property (unless you figured out it won't be necessary for some reason... which is quite possible).

application/console/migrations/m170203_210542_create_initial_tables.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+use yii\db\Migration;
+
+class m170203_210542_create_initial_tables extends Migration
+{
+    public function up()
```
forevermatt: If we use `safeUp()` and `safeDown()` (instead of simply `up()` and `down()`) then it will wrap these in a transaction, and Scrutinizer won't complain about method names being too short. :-)

application/console/migrations/m170203_210542_create_initial_tables.php
```diff
@@ -0,0 +1,75 @@
...
+                'user_id' => 'int(11) not null',
+                'password_hash' => 'varchar(255) not null',
+                'created_utc' => 'datetime not null',
+            ],
+            "ENGINE=InnoDB DEFAULT CHARSET=utf8"
+        );
+
+        $this->createIndex('idx_password_hash','{{password_history}}','password_hash',false);
+
+        $this->addForeignKey('fk_password_history_user_id','{{password_history}}','user_id','{{user}}','id','CASCADE','NO ACTION');
```
forevermatt: For some reason I thought it was better to use `'CASCADE'` for both update and delete. Was there a specific reason you left update as `'NO ACTION'`?

application/features/authentication/authentication.feature
```diff
@@ -0,0 +1,30 @@
...
+
+  Scenario: Create an authentication for a known user with a matching password
+    Given I am authorized
+      And I receive a username
+      And I receive a password
+    When I verify the password matches the associated password
+    Then I will respond positively
+
+  Scenario: Attempt to create an authentication for an unknown user
+    When I realize the username does not exist in my system
```
forevermatt: You might want some number of `Given`s before this `When`... perhaps something like this:
```
Given I am authorized
  And I receive a username
  And I receive a password
  But that username does not exist in my system
When I attempt to verify the username and password
Then I will respond negatively
```
Thoughts?

application/features/authentication/authentication.feature
```diff
@@ -0,0 +1,30 @@
...
+    Then I will respond negatively
+
+  Scenario: Attempt to create an authentication without a username
+  Scenario: Attempt to create an authentication with an invalid username
+
+  Scenario: Attempt to create an authentication for a known user with a mismatched password
+    When I realize the given password has does not match the stored version
```
forevermatt: As above, you may want a `Given` or two to establish the state.

application/common/config/main.php
wcjr: Thanks for tracking this:  https://github.com/silinternational/php-env/issues/1

application/common/helpers/Utils.php
wcjr: I agree with that sentiment of the code smell, I'll make a change for that.

application/features/password/password.feature
```diff
@@ -0,0 +1,37 @@
...
+  Scenario: Provide a new password for an existing user
+    Given I receive an existing employee id
+      And the requestor is authorized
+      And the user does not already have a password
+      And I receive a password
+    When I receive a request to create a password for a specific user
+    Then a new password hash should be stored
+      And the last changed date should be stored as the instant it was stored
+      And the last changed date should be stored in UTC
+      And the last synched date should be stored as the instant it was stored
```
forevermatt: This came up in a comment above, but I don't think the last-synced date is supposed to change when a password is set. @fillup, is that right?

application/common/helpers/Utils.php
wcjr: https://github.com/silinternational/idp-id-broker/commit/29ac0a3512b34eb3d0bf28381c1a974eb41fad09

application/common/models/PasswordHistory.php
wcjr: yes, probably a worthwhile tweak.  The current implementation is relying on the fact that `created_utc` won't be provided but is required...it's easy enough to supplement that with a behavior.  The default will still be needed as well. https://github.com/silinternational/idp-id-broker/commit/8538d733d2160abb37b8ee405a38c988aa5fc118

application/common/models/User.php
wcjr: I'm certainly not married to this implementation and appreciate the distinction of this sort of validation since there's no actual update to the database occurring.  It can be very tempting to utilize aspects of a framework to accomplish some task in a way that is out of scope or never intended by the authors...you know, the old "once I have a hammer everything starts looking like a nail."

I'm attempting to stick to a pattern I see emerging from the Yii docs that advocates this idea of strict adherence to the MVC architecture (which I realize can even have its own varied interpretations).  For everything so far, I rely on the model to establish all the "rules" whether or not they're related to data integrity like types or business rules like emails need to be a "valid".  The semantics of my thinking here is, "the password needs to match the hash" is similar to "the email needs to match an acceptable structure".  In my mind I made the semantical argument that a rule is rule and the model knows best how to ensure its integrity.

In the end, I believe this will boil down to stylistic choice and am comfortable refactoring it in a way that is more comfortable for you and @fillup if he feels the same way.  I'm assuming a refactoring would simply result in an explicit method call rather than relying on Yii's existing lifecycle, does that sound about right?

application/common/models/User.php
wcjr: it's purely based on the order of their declaration...I've found so many of Yii's conventions rely on the ordering of the declarations that this didn't sit outside of that pattern enough to warrant additional safeguards.  The test cases would catch any unintended consequences of code changes here.

application/common/models/User.php
wcjr: Ahhh, that's an excellent consideration.  I have been viewing the sync more broadly representing the idea of the fact that the information is accurate as of some point, not solely a field that should only be affected by our idsync module...if that's the case we will need to ensure we can distinguish POST's based on consumer so we know not to update the `last_synced` column when a request has come in to update user info, e.g., user mgmt...I think that's going to require some additional spec.  At this point that might be as simple as assuming that _only_ POST's to `/user` affect that `last_synced_utc` column under the assumption the only consumer doing that is the id-sync, sound right?

application/common/models/User.php
wcjr: agreed, I'm not aware of their specific use to a consumer yet either.  I could go ahead and remove them until we find a need for them...are you ok with that approach until we know more?

application/features/user/user.feature
```diff
@@ -0,0 +1,196 @@
...
+      And the property message should contain "not found"
+      And a user was not created
+
+    Examples:
+      | action    |
+      | retrieved |
+      | updated   |
+      | deleted   |
+
+  Scenario Outline: Attempt to create a new user without providing a required property
```
forevermatt: Very nice... I like you you compacted so many tests into a single scenario in such an incredibly readable way. Kudos, @wcjr !

application/features/user/user.feature
```diff
@@ -0,0 +1,196 @@
...
+    Then a user exists with an employee_id of "123"
+      And the only property to change should be last_synced_utc
+      And last_synced_utc should be stored as now UTC
+
+  Scenario Outline: Change the properties of an existing user
+    Given the requester is authorized
+      And there are no users yet
+      And I provide all the required fields
+      And I request the user be created
+      And a user exists with an employee_id of "123"
```
forevermatt: I'm not sure I understand the `Given`s (aka. state) here. It looks like there has to be both "no users" and "a user with a specific employee_id". I wonder if it would be worth refactoring this set of `Given`s a little to reduce that apparent contradiction.

application/common/models/User.php
wcjr: only if you notice from the caller that this method is only called in the `UPDATE_PASSWORD` scenario.  The `Model` always determines which attributes are saved or not based on whether the attribute is active or not and an attribute's dirtiness...this was always happening whether scenarios were in the mix or not.  Am I understanding your concern here as being the lack of a attribute assignments close to the save?

application/frontend/config/main.php
```diff
@@ -2,64 +2,46 @@
...
             'rules' => [
-
                 /*
-                 * Status route
+                 * Status
                  */
                 'GET /site/system-status' => 'site/system-status',
 
                 /*
-                 * Catch all to throw 401 or 405
```
forevermatt: Re: our conversation about throwing a 401 vs. a 404 for an unauthorized request to a non-existent endpoint, this may be an example of how to do that.

application/common/models/User.php
wcjr: I believe the "rule" for determining the validity of the data here is the `PasswordHistory`'s concern...and, the fact that I can't simply verify whether two hashes are equal or not requires the PasswordHistory have the `password` to utilize php's `password_verify` call.  Is there a way to determine two hashes are equal?  If so, the `PasswordHistory` will *not* need the `password`.

application/frontend/controllers/AuthenticationController.php
```diff
@@ -0,0 +1,38 @@
...
+     * @throws HttpException
+     */
+    public function actionCreate(): User
+    {
+        $user = User::findOne([
+            'username' => \Yii::$app->request->getBodyParam('username')
+        ]);
+
+        if ($user === null) {
+            throw new NotFoundHttpException();
```
forevermatt: I think this detail already came up, but for the sake of documenting it for later discussion, we'll want to refactor this per our auth-sequence diagram.

application/frontend/controllers/SiteController.php
```diff
@@ -2,83 +2,33 @@
...
-                        'roles' => ['?'],
-                    ],
-                ]
-            ],
-            'authenticator' => [
-                'except' => ['system-status'] // bypass authentication for /site/system-status
-            ]
-        ]);
-    }
+        $behaviors = parent::behaviors();
```
forevermatt: @fillup Do you have any thoughts on using this approach (get the parent values, make changes, return the result) versus our typical ArrayHelper::merge approach? I think consistency would suggest we keep using merge, but for someone who isn't familiar with the details of how Yii's array merge function works, the get-change-return approach is clearer.

application/frontend/controllers/UserController.php
```diff
@@ -0,0 +1,44 @@
...
+{
+    public function actionCreate(): User
+    {
+        $existingUser = User::findOne([
+            'employee_id' => Yii::$app->request->getBodyParam('employee_id')
+        ]);
+
+        $user = $existingUser ?? new User();
+
+        $user->attributes = Yii::$app->request->getBodyParams();
```
forevermatt: I'm sorry to bring this up again (and maybe it's just a mindset shift I need to work through), but I still cringe at using massive assignment to collect data from the request and save it to the database. It feels like we're completely tossing out the opportunity to effectively do a whitelist for what properties we are willing to accept from requests. Does that seem reasonable, or am I just obsessing over something that doesn't matter?

application/frontend/controllers/UserController.php
```diff
@@ -0,0 +1,44 @@
...
+            'employee_id' => $employeeId
+        ]);
+
+        if ($user === null) {
+            throw new NotFoundHttpException();
+        }
+
+        $user->scenario = User::SCENARIO_UPDATE_PASSWORD;
+
+        $user->attributes = Yii::$app->request->getBodyParams();
```
forevermatt: (Same as above)

application/common/models/UserBase.php
wcjr: I like that!  https://github.com/silinternational/idp-id-broker/commit/2c1302ce364ffd3fd7780dac4e2d3316dc3d3de2

application/frontend/web/.htaccess
```diff
@@ -0,0 +1,12 @@
...
+# if request was not https, redirect to https, if env var is present
+RewriteCond %{HTTP:X-Forwarded-Proto} ^http$
+RewriteRule (.*) https://%{HTTP_HOST}%{REQUEST_URI} [R=301,L]
+
+# If a directory or a file exists, use the request directly
+RewriteCond %{REQUEST_FILENAME} !-f
+RewriteCond %{REQUEST_FILENAME} !-d
+
+# Otherwise forward the request to index.php
+RewriteRule . index.php
```
forevermatt: It's surprising to me that this `.htaccess` file wasn't already in the repo. Are these rules normally something we specify in the apache conf file instead?

application/console/config/main.php
wcjr: if that's the case, I'll add it back it in.  For now, it's not needed on migrate or Gii calls so I removed it.

application/frontend/web/load-app.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+require(__DIR__ . '/../../vendor/autoload.php'); //TODO: why is this required for the web app but not the tests?
```
forevermatt: Are you referring to PHPUnit tests? If so, I think it's this: By the time the tests are run, the bootstrap file for PHPUnit has already pulled in the autoloader (otherwise it couldn't have loaded the classes used for running the tests). For web access, this is the first file run, so any use of classes needs to have the autoloader already in place.

application/console/migrations/m170203_210542_create_initial_tables.php
wcjr: much better, thanks:  https://github.com/silinternational/idp-id-broker/commit/d98b70f64a696255cdd4c405c2bd672f51bfd651

application/frontend/web/index.php
```diff
@@ -1,16 +1,5 @@
 <?php
-defined('YII_DEBUG') or define('YII_DEBUG', true);
-defined('YII_ENV') or define('YII_ENV', 'dev');
 
-require(__DIR__ . '/../../vendor/autoload.php');
-require(__DIR__ . '/../../vendor/yiisoft/yii2/Yii.php');
-require(__DIR__ . '/../../common/config/bootstrap.php');
-require(__DIR__ . '/../config/bootstrap.php');
+$application = require('load-app.php');
```
forevermatt: Unfortunately, I think we may need to keep all this code here in `index.php` rather than another file here in the `web` folder, because people could directly access the new `web/load-app.php` file instead. What was the thought for pulling this out? Was it for getting Yii stuff available to the Behat tests?

application/console/migrations/m170203_210542_create_initial_tables.php
wcjr: no, didn't even question it when I copy-pasted it but I agree with you.  https://github.com/silinternational/idp-id-broker/commit/e6999f12ca1210259b83b90668d9291aae907cd2

application/features/authentication/authentication.feature
wcjr: most definitely, sorry I didn't make it clear that I hadn't gotten to these at all yet...it was simply a scratchpad for thoughts at the time.  These will be built out in https://github.com/silinternational/idp-id-broker/issues/7

application/features/authentication/authentication.feature
wcjr: will be done in https://github.com/silinternational/idp-id-broker/issues/7

application/features/user/user.feature
wcjr: glad that was helpful, I know it certainly saved me some time once I got the hang of it!

application/run-tests.sh
```diff
@@ -1,9 +0,0 @@
-#!/usr/bin/env bash
```
forevermatt: As a side note, in some projects I have ended up moving back to having a `run-tests.sh` file so that the way I run tests locally is the same as the way tests are run on CI/CD.

application/features/user/user.feature
wcjr: yeah, excellent idea, e.g., there only exists one user with an `employee_id` of "123"

application/frontend/config/main.php
wcjr: right, thanks!

application/frontend/controllers/AuthenticationController.php
wcjr: yes, thank you, capturing a reminder for that in https://github.com/silinternational/idp-id-broker/commit/830723a2f7970f34f8412cca6f0dba503d9c355f

application/frontend/controllers/SiteController.php
wcjr: yep, happy to use the merge everywhere if that's what you all prefer.

application/frontend/controllers/UserController.php
wcjr: It's totally reasonable, I completely understand where you're coming from.  It was a little too "magicky" for me when I first saw it but I came around to liking the idea, you know how some designs grow on you once you've begun using and appreciate where the authors were coming from?  FWIW, there is still a whitelisting process, it's just be done via scenarios instead of through getters in the controller.   This aversion you're having may simply be about the notion of who's the main actor here, the controller or the model...let's see how it strikes @fillup , if the style is a no-go for you all because it's too different from what you want to see, I'll refactor.

application/frontend/web/.htaccess
wcjr: oh, I don't know about that... @fillup ?

application/frontend/web/load-app.php
wcjr: not PHPUnit, but Behat.  I assume the same applies though.  I'd like to turn off that autoloader for the tests if possible then, is that controlled through composer somehow?

application/frontend/web/index.php
wcjr: ok...I was pulling out what was necessary for both the webapp and the tests.  This also allowed me to understand what exactly it was that the webapp needed that was different from the tests.  I was hoping to make them both the same.

application/run-tests.sh
wcjr: Could the `make test` suffice for that as well?

application/common/models/User.php
wcjr: went ahead and removed them for now: https://github.com/silinternational/idp-id-broker/commit/967090a340e3bd95a8df73e36bfd4f7ab20ceec3

application/common/models/User.php
forevermatt: Yeah, and explicit call to some new method on User is what I was thinking.

application/features/bootstrap/UserContext.php
```diff
@@ -44,7 +44,7 @@ public function theUserStoreIsEmpty()
     }
 
     /**
-     * @When /^I request the user be (created|updated|deleted|retrieved|<.*>)$/
+     * @When /^I request the user be (created|updated|deleted|retrieved|headed|patched|<.*>)$/
```
forevermatt: Does `<.*>` actually match other words, or would it need to simply be `.*`?

application/features/bootstrap/UserContext.php
```diff
@@ -72,18 +72,22 @@ private function buildClient(): Client
...
+            default: throw new InvalidArgumentException("$action is not a recognized HTTP verb.");
         }
     }
 
     private function extractBody(Response $response): array
     {
         $jsonBlob = $response->getBody()->getContents();
 
-        return json_decode($jsonBlob, true);
+        return json_decode($jsonBlob, true) ?? [];
```
forevermatt: It looks like this will silently ignore failures to decode the JSON string. Is that the intent?

application/features/user/user.feature
```diff
@@ -3,27 +3,38 @@ Feature: User
   As an authorized user
   I need to be able to manage user information
 
-  Scenario: Attempt to create a user as an unauthorized user
+  Scenario Outline: Attempt to act upon a user as an unauthorized user
```
forevermatt: Do you think it would be worth using a different word for the user (record) and for the user (initiating the HTTP request)?

application/features/bootstrap/UserContext.php
wcjr: That's a great point...I simply did that to match the angled brackets for use in Scenario Outlines...maybe there's a better regexp for it, something to "optionally" match the angled brackets.

application/features/bootstrap/UserContext.php
wcjr: The intent was to allow for 204's (where there's no content).  I think JSON syntax issues will come up all over the place if that's the case.

application/features/user/user.feature
wcjr: yeah, that would be better.  Thanks.

application/run-tests.sh
forevermatt: Unfortunately, no, because the tests are run by Codeship as a docker container, without the help of Make.

wcjr: ah, I didn't realize that.  Ok, I'll add it back in when we get to CI then.  Thanks.

application/features/bootstrap/UserContext.php
forevermatt: What if you only tried to decode as JSON if the body's contents were non-empty? You could simply return an empty string if the body was empty.

application/console/config/main.php
fillup: `log` should be bootstrapped in common and I think you need `gii` in console for rebuilding base models. Pretty sure we need `controllerNamespace` for a `CronController` to be called from CLI.

application/features/password/password.feature
fillup: Correct, `last synced` only updates when updated via a sync process.

fillup: Another option/considering is to sync the user anytime they change their password. That could provide a mechanism for forcing an individual sync.

application/frontend/controllers/SiteController.php
fillup: I'm fine this way.

application/frontend/controllers/UserController.php
fillup: As long as we have unit/api tests to ensure unsafe attributes cannot be set I'm okay with it. But we need multiple tests for each scenario to be sure it works properly.

application/frontend/web/.htaccess
fillup: No, I think probably when the template repo got created it was missing or something.

application/frontend/web/index.php
fillup: Yeah I think moving it back into index is probably better. Otherwise should add rule to .htaccess to prevent accessing load-app.php and maybe a check inside that file itself to make sure it is not being requested directly. Probably easier just to bring back into index.

application/features/bootstrap/UserContext.php
wcjr: Good idea, returns an empty array though.  https://github.com/silinternational/idp-id-broker/commit/ca711632db335b57c11c79e19ab793d4789c257a.

Too bad `json_decode` doesn't throw a parse error or something :-(

application/common/models/User.php
wcjr: no, not sure what happened :-)  https://github.com/silinternational/idp-id-broker/commit/bca9597e9dc34d7a2a508c9a8db1f841f1f1bc83

application/console/config/main.php
wcjr: yep, you're right @fillup , I did end up needing `gii` for the basemodels.  I'll bring all the necessary config's back in as we need them, e.g., `CronController`

application/features/password/password.feature
wcjr: ok, sounds like something I'd like to talk through a bit more...especially considering some of the redesign @forevermatt and I have been contemplating around `POST` and `PUT`.  Adding a reminder:  https://github.com/silinternational/idp-id-broker/commit/53ee077cde3d9bbc0f1de4898545198d4ecd756b

application/frontend/controllers/SiteController.php
wcjr: ok, thanks

application/frontend/controllers/UserController.php
wcjr: We're definitely covered on all permutations I could think of.

application/features/bootstrap/UserContext.php
forevermatt: @wcjr Yii adds a class for doing exactly that:
http://www.yiiframework.com/doc-2.0/yii-helpers-json.html

application/frontend/web/index.php
forevermatt: Moving it to a different folder would solve the problem of it being web-accessible, but I think I would lean towards simply duplicating the relevant pieces, because the tests are inherently separate from the application itself, so it seems fine that they each have a copy of this, especially since they need to be different. I hit a similar situation in SilAuth of needing to initialize a Yii application but not `run()` it. Thoughts?

application/common/models/User.php
wcjr: https://github.com/silinternational/idp-id-broker/issues/27

application/features/password/password.feature
wcjr: https://github.com/silinternational/idp-id-broker/issues/27

application/frontend/web/index.php
wcjr: I don't want to have to worry about keeping those files in sync.  I refactored this a bit to eliminate that visibility issue and deal with whether to `run()` or not to the initiator.  https://github.com/silinternational/idp-id-broker/commit/f6f89e36cd91783e90e823199621b2fdc2ae0edb

application/features/bootstrap/UserContext.php
wcjr: https://github.com/silinternational/idp-id-broker/commit/94e28c4ed07525f91bfa514eb0a2ca9dc9533cce

application/features/bootstrap/UserContext.php
wcjr: Hugh D Mann!!!  Thank you: https://github.com/silinternational/idp-id-broker/commit/eee6a87ac9de3c61b53f38ac464edf5bd1e2dec0

application/common/models/User.php
forevermatt: I think my concern is that, if this `savePassword(...)` method was called in a scenario other than `UPDATE_PASSWORD`, it would save more than the password. Based on it's name, I would think we'd want to remove the `$attributeNames` parameter to this `savePassword` function and here in the call to `parent::save(...)` specifically instruct it to only save the `password` (and/or `password_hash`... I'm not sure which). It's one of those situations where it feels like we're saying "oh, some other piece of code will enforce the proper constraints for this function" rather than having the function only do what it's name claims it will do.

Thoughts?

application/common/models/User.php
```diff
@@ -0,0 +1,185 @@
...
+
+        return parent::save($runValidation, $attributeNames);
+    }
+
+    private function savePassword($runValidation = true, $attributeNames = null): bool
+    {
+        $transaction = ActiveRecord::getDb()->beginTransaction();
+
+        try {
+            if ($this->hasPasswordAlready()) {
```
forevermatt: Is this code still just trying to save the previous password to the password history list, or has it been updated to immediately save the new password to the password history list?

application/common/models/User.php
wcjr: tightening that up is better, thanks:  https://github.com/silinternational/idp-id-broker/commit/601072bae092ce6c1c23b630ccfaf82c6d77148d

application/common/models/User.php
wcjr: line 141 will save the password history, i.e., it will save the current hash to the history table, line 148 will save the new password hash, i.e., it save the new hash to the user table...if both work then the transaction will be committed on line 154.  Sounds like I need some more descriptive methods or some comments?

forevermatt: I guess I don't understand why we're only trying to save the password history if `hasPasswordAlready()` returns true. Shouldn't we always try to `saveHistory()`?

wcjr: `password_hash` is nullable.  To my understanding, we do NOT need a password in order to create a user.  If that's true, then we only need to create password history in the event the user already has a password.

forevermatt: Ohhh... I hadn't even thought of that. But isn't this code only called when we are trying to save a password?

wcjr: yes :-)  The scenario still exists where a user is trying to change their password...from _nothing_ to something in this case.  Want to work through those semantics?

PR #16, opened 2017-03-22
Refactor api.raml to use RAML 0.8.

forevermatt:


forevermatt:
@wcjr or @fillup, would you glance at this and see if you see anything that needs fixed?

PR #22, opened 2017-03-24
Revise API spec

forevermatt:


api.raml
```diff
@@ -104,11 +208,55 @@ schemas:
...
           authentication failed).
+        body:
+          application/json:
+            schema: Error
+            example: |
+              {
+                "name": "Bad Request",
+                "message": "Some error message about an invalid login.",
+                "code": 0,
+                "status": 400
```
wcjr: This is still pending discussion correct?  It's currently considered a 422.

api.raml
```diff
@@ -122,16 +270,72 @@ schemas:
...
+        description: The new user record was created.
       422:
         description: The given user data does not satisfy some validation rule.
+        body:
+          application/json:
+            schema: Error
+            example: |
+              {
+                "name": "Unprocessable entity",
+                "message": "Employee ID cannot be blank.",
```
wcjr: This message probably needs to be more generic since it can have lots of reasons.

api.raml
```diff
@@ -69,6 +92,87 @@ schemas:
         },
         "type": "object"
       }
+  - UserList: |
+      {
+        "$schema": "http://json-schema.org/draft-03/schema",
+        "items": {
+          "properties": {
+              "employee_id": {
+                "required": false,
```
wcjr: would this mean the default list would not contain any properties?

forevermatt: No, just that this type of data is not require to have any particular properties. The place this is currently used (`GET /user`) defaults to include `employee_id` and `active`. It is possibly to request a specific set of fields that doesn't include either of those two defaults, though, so even they aren't required for this type of data.

api.raml
forevermatt: Good catch. I'm fine with `422` for this. I'll go ahead and change it unless you object.

api.raml
wcjr: ah, gotcha

api.raml
forevermatt: Yeah, I'm not sure what to do here. Since this is just an example, it seems more helpful to have a realistic example, even if it only represents one possible response, rather than a more generic example that will never actually be returned. Thoughts?

wcjr: Since it's just an example, it should be fine.  I didn't consider the fact that it's just one of hundreds of permutations.

PR #28, opened 2017-03-29
Return user info on successful user-create.

forevermatt:


api.raml
```diff
@@ -273,6 +273,22 @@ schemas:
...
+        body:
+          application/json:
+            description: The newly-created user record.
+            schema: User
+            example: |
+              {
+                "employee_id": "12345",
+                "first_name": "John",
+                "last_name": "Smith",
+                "display_name": "John Smith",
```
forevermatt: @fillup Will we generate a `display_name` (either on user creation or dynamically on returning a user's info) even if none was given?

wcjr: not at this point...shall we?

fillup: I dont think we need to, but I also dont see harm in doing it if null.

wcjr: the code to accomplish that is easy enough, @forevermatt let me know what you end up deciding here.

forevermatt: I guess it seems better to have the ID Broker generate the display name (when it's requested and the field is `null`) rather than having each app that gets info from the ID Broker having to separately duplicate that logic of how to generate a display name when not given one.

wcjr: you got it. https://github.com/silinternational/idp-id-broker/issues/39

fillup:
Ready to merge? Or are you going to implement actual support for returning full user object on create as well?

forevermatt:
I was just updating the spec at this point.

Issue #20, opened 2017-03-23
Obfuscate any differences between failed authentications


forevermatt:
We'll want to make sure that calls to `POST /authenticate` that are not acceptable (for whatever reason) all look identical. In other words, whether the username is wrong (i.e. no such user), the password is wrong, the user is inactive, or the user is locked, the request should take approximately the same amount of time to return and the response should be identical for all of those scenarios.

wcjr:
relevant code:  

https://github.com/silinternational/idp-id-broker/blob/feature/initial-features/application/frontend/controllers/AuthenticationController.php#L25-L37

and

https://github.com/silinternational/idp-id-broker/blob/feature/initial-features/application/common/models/User.php#L83

fillup:
Currently the errors are different if the user doesn't exist versus if the password was incorrect. The error displayed to the end user needs to be the same, but do we want it to be the same from the broker to auth app, or does it help auth app to know the difference between exists/doesn't and incorrect pw?

forevermatt:
I don't think the auth. app gains anything by knowing that. In face, it's helpful if it doesn't know.

When checking a username/password combination, from a security standpoint we want there to be (as much as possible) no discernible difference for the requester between a non-existent user, a wrong password, a disabled user, etc.

The easiest/simplest way I see to do that is to not expose that difference to any part of the system. Since successfully concealing those differences takes some careful design, hiding the complexities of that design in a single place in a single piece of the system (the ID Broker, ideally in a single function somewhere like the `User` class) seems like it would be good approach.

wcjr:
@forevermatt @fillup FYI, I'm going to leave the error message blank and will also ensure even system, runtime errors come out as a generic bad request, i.e., those that might've been 500 will also come out as a 400.

forevermatt:
@wcjr Okay. At the moment, I would think that (other than "acceptable credentials" and "unacceptable credentials"), the one other useful response would be "something went wrong, I don't know whether the credentials were acceptable" (meaning that 500 errors would be distinguishable from 400 errors). I'll leave that to you for now, though, since you're obviously much more in the middle of that context than I am.

wcjr:
np, easy enough to distinguish them.

forevermatt:
Since part of this issue (preventing timing attacks) is going to be addressed through a later change, do you mind if I re-open this until that additional piece is also done?

wcjr:
[...]

PR #30, opened 2017-03-30
Feature/authentication tests

### LINKED ISSUE #7: Build automated tests for the authentication feature

wcjr:
resolves #7 .

If @fillup and @forevermatt are not interested in reviewing this, let me know so I can go ahead and merge it.


application/features/authentication.feature
```diff
@@ -0,0 +1,135 @@
+Feature: Authentication
+  In order to identify a specific user
+  As an authorized user
```
forevermatt: Very minor point, but this `user` could probably be changed to `requester`.

application/features/authentication.feature
```diff
@@ -0,0 +1,135 @@
...
+        | active       | yes                   |
+        | locked       | no                    |
+
+  Scenario: Attempt to authenticate an unknown user
+    Given I provide the following valid data:
+        | property  | value     |
+        | username  | daddy_o   |
+        | password  | govols!!  |
+    When I request "/authentication" be created
+    Then the response status code should be 404
```
forevermatt: According to [our current spec](https://github.com/silinternational/idp-id-broker/blob/develop/api.raml#L207), I think the response for unacceptable credentials would be a `422`. I know some of that needs/needed to be refactored, but it might be worth going ahead and writing the tests to match the current spec. Thoughts?

forevermatt: Ditto for the various other failed authentication calls below.

application/features/authentication.feature
wcjr: Thanks, that's better:  https://github.com/silinternational/idp-id-broker/commit/6365259ec1b19225759bcce43b1a8f09096e6a2a

application/features/authentication.feature
wcjr: totally, agreed, I want to keep all that work in #20

application/features/authentication.feature
```diff
@@ -0,0 +1,135 @@
...
+      | username | 21    |
+
+  Scenario: Attempt to authenticate a known user with a mismatched password
+    Given I provide the following valid data:
+      | property  | value      |
+      | username  | shep_clark |
+      | password  | govols     |
+    When I request "/authentication" be created
+    Then the response status code should be 400
+      And the property message should contain "password"
```
forevermatt: It seems better to simply define the desired result to be that the authentication was not successful (then leave the details of what "not successful" means to the implementation in the Behat context file).

application/features/authentication.feature
forevermatt: Okay, thanks.

application/features/authentication.feature
wcjr: I suppose then you would suggest the same refactoring for all those user tests as well?

forevermatt:
@wcjr Is there anything in particular you want us to look at? It mostly looks like simply rearranging/combining some files, so I didn't examine much of it. If you're not itching for feedback on anything in particular, I'm fine with you merging it.

application/features/authentication.feature
forevermatt: I don't think so, since in the `/user` calls the ID Broker API returns much more (and more varied) information, and it might be worth making sure that is all correct. With authentication, by design we want it to only be able to return one of two possible results (barring `5xx` server errors). I think that's all I was getting at.

wcjr: ok, I'll make sure those are refactored when I work on #20

PR #33, opened 2017-03-30
sync date was updating incorrectly, resolves #27

### LINKED ISSUE #27: Ensure `last_synced_utc` is updated appropriately
This field should only ever be updated when a sync has occurred, not password changes.  At this time, that might mean only when a `POST` has occurred.

wcjr:


application/common/models/User.php
```diff
@@ -145,7 +151,7 @@ private function savePassword($runValidation = true): bool
 
             $this->password_hash = password_hash($this->password, PASSWORD_DEFAULT);
 
-            if (! parent::save($runValidation, ['password_hash'])) {
+            if (! parent::save()) {
```
forevermatt: Was the `['password_hash']` argument unnecessary after all?

application/features/password.feature
```diff
@@ -37,3 +37,5 @@ Feature: Password
 #  Scenario: Attempt to delete a password
 #  Scenario: Attempt to change the password of an existing user using the same password they already have.
 #  Scenario: consider invalid employee ids that test the type conversions of Yii and/or PHP, e.g., /user/[true|false|1|0]/password
+#  Scenario: ensure last_synced_utc does NOT change when changing a password.
+#  Scenario: ensure last_changed_utc does change when changing a password.
```
forevermatt: It might also be worth adding a scenario to ensure that no other attributes (beyond `last_changed_utc`) can be changed by a call to set the password.

application/common/models/User.php
wcjr: no

application/features/password.feature
wcjr: good idea.

application/common/models/User.php
forevermatt: Umm... I'm sorry. I'm not sure which way that "no" means, but since you're planning on adding a test to catch what I was concerned about here (possibly changing other fields through a call to `savePassword()`), I'll assume you've got that taken care of.

wcjr: yes, I can understand that..., sorry.  I meant no, it was not necessary.  This continues to go back to the idea of the "active" attributes involved in the scenario.  There's no way for any other attributes to make it into the mix.  It's certainly a worthwhile test to have though so I'm putting a reminder in place for it.  https://github.com/silinternational/idp-id-broker/commit/b794df9a4b8a235de5405e4f63980cc189843027

PR #36, opened 2017-03-30
Feature/obfuscate failed authn

### LINKED ISSUE #20: Obfuscate any differences between failed authentications

wcjr:
resolves #20

### DIFF
--- a/api.raml
+++ b/api.raml
@@ -204,7 +204,7 @@ schemas:
                 "active": "yes",
                 "locked": "no"
               }
-      422:
+      400:
         description: |
           The given credentials were NOT acceptable/valid (i.e. - the
           authentication failed).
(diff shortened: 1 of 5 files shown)


wcjr:
so sorry about the commit message on commit 2349fe2...just saw that.  That commit was related to incorporating code review feedback about test semantics.

wcjr:
thank you @forevermatt , fixed it:  https://github.com/silinternational/idp-id-broker/commit/739ad5c62a1a37fcfda2a72dcf57ab8e03cab16d

PR #42, opened 2017-04-03
Default display_name if null to first + last

fillup:


fillup:
first and last are required, so I don't think accounts will get into broker without one or the other

wcjr:
oh yeah.

fillup:
I made those edits via web interface though so test them out to make sure I didnt put garbage in there

PR #46, opened 2017-04-06
was not able to identify one consumer from another, working on #43

### LINKED ISSUE #43: Auth tokens need to be specific to consumer
Each consumer needs a unique token.

wcjr:
resolves #43

### DIFF
--- a/application/common/models/ApiConsumer.php
+++ b/application/common/models/ApiConsumer.php
@@ -8,11 +8,18 @@ class ApiConsumer extends Component implements IdentityInterface
 {
     public static function findIdentityByAccessToken($token, $type = null)
     {
-        if ($token === getenv('API_ACCESS_KEY')) {
+        if (self::isRecognized($token)) {
             return new ApiConsumer();
         }
 
         return null;
+    }
+
+    private static function isRecognized($token): bool
+    {
+        $recognizedKeys = explode(',', getenv('API_ACCESS_KEYS'));
+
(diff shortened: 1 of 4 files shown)


forevermatt:
It's kind of funny, this is almost exactly what I ended up doing in ID Sync:
https://github.com/silinternational/idp-id-sync/blob/develop/application/common/components/ApiConsumer.php

wcjr:
funny :-)  

I suppose I need to understand the significance of forcing `true` in the array check?  What use case are you protecting against there?

Additionally, should I be using `params` too?

forevermatt:
True is to force a strict comparison (checking type specifically, rather than allowing type-casting) so that something like `false` isn't considered to match an empty string (`''`), which otherwise could happen:
```
> php -r "var_dump(in_array(false, ['']), in_array(false, [''], true));"
bool(true)
bool(false)
```
I don't know whether there's actually a chance of hitting that here, but using `true` for the strict comparison removes the risk.

forevermatt:
As far as using `params`, it just felt like it was more in line with the way Yii expects you to provide it configuration details (rather than grabbing them from the environment variables directly within your normal code, almost like some global variable). Not mandatory, but it seems like a better practice.

wcjr:
gotcha, I'll go back and refactor those then...I also noticed now that I didn't use the Sil version of the env getter.

wcjr:
Thanks again @forevermatt , took care of these in #53

PR #48, opened 2017-04-10
Update tests to use first entry in API_ACCESS_KEYS.

forevermatt:


forevermatt:
The local `make test` run failed, I think related to the change from because API_ACCESS_KEY to API_ACCESS_KEYS... but I may have been wrong. I'm still finding environment variables defined various places. :-)

forevermatt:
Is the `local.env.dist` file up-to-date according to what's in your `local.env` file?

wcjr:
That shouldn't be the case, I'll give you a call.

PR #50, opened 2017-04-13
Lazy-load passwords using LDAP

forevermatt:


forevermatt:
@wcjr Good catch. I forgot to add a toggle-flag in the configuration.

forevermatt:
@wcjr Okay, the feature toggle is present now. Would you mind glancing at this again?

PR #57, opened 2017-04-19
Feature/password expiration

### LINKED ISSUE #55: Password expiration
broker should calculate a password expiration datetime (UTC) and return in a `password_expires_at_utc` property in the `User` object.  Passwords should expire 1 year from the datetime of the last password change.  Configurability will come later if needed.

@Baggerone @forevermatt

wcjr:
resolves #55

### DIFF
--- a/application/common/models/PasswordHistory.php
+++ b/application/common/models/PasswordHistory.php
@@ -58,6 +58,11 @@ class PasswordHistory extends PasswordHistoryBase
         return false;
     }
 
+    public function expires()
+    {
+        return MySqlDateTime::format(strtotime('+1 year', strtotime($this->created_utc)));
+    }
+
     public function behaviors(): array
     {
         return [
--- a/application/common/models/PasswordHistory.php
+++ b/application/common/models/PasswordHistory.php
@@ -43,10 +43,12 @@ class PasswordHistory extends PasswordHistoryBase
 
     private function hasAlreadyBeenUsedWithinLimit(): bool
     {
+        $reuseLimit = Yii::$app->params['passwordReuseLimit'];
+
         /** @var PasswordHistory[] $previousPasswords */
         $previousPasswords = $this->user->getPasswordHistories()
                                         ->orderBy(['id' => SORT_DESC])
-                                        ->limit(10)
+                                        ->limit($reuseLimit)
                                         ->all();
 
         foreach ($previousPasswords as $previousPassword) {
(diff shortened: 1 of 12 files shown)


application/common/models/PasswordHistory.php
```diff
@@ -58,6 +58,11 @@ private function hasAlreadyBeenUsedWithinLimit(): bool
         return false;
     }
 
+    public function expires(): string
+    {
+        return MySqlDateTime::format(strtotime('+1 year', strtotime($this->created_utc)));
```
fillup: Can we move the `+1 year` to a config parameter and then pull from env with a default of `+1 year`? Different IdPs may want different policies.

application/common/models/User.php
```diff
@@ -189,17 +194,45 @@ public function rules(): array
...
+        /** @var $mostRecentPassword PasswordHistory */
+        $mostRecentPassword = $this->getPasswordHistories()
+                                   ->orderBy(['id' => SORT_DESC])
+                                   ->one();
+
+        return $mostRecentPassword->expires();
+
+    }
+
+    private function addGracePeriod(string $expiration): int
```
fillup: Can we confirm with Steve that this method of adding the grace period will still allow the IdP to force redirection to change password rather than logging in? The idea is 14 days before expiration it will give a warning. After expiration, up to the grace period after logging in they will be prompted/forced to go change password before logging in further. I'm assuming looking at this that it just extends their password expiration by 30 days to allow successful authentication but I'm not sure if the IdP will be aware that it is "expired".

application/common/models/User.php
```diff
@@ -189,17 +194,45 @@ public function rules(): array
...
+    {
+        return strtotime('+30 days', strtotime($expiration));
```
fillup: Same as lifetime above, this should be an app config param, retrieved from an env var with default of `+30 days`

application/common/models/PasswordHistory.php
wcjr: certainly, we discussed that option and thought it was prudent at this point to hold off until there was request for it.  If you like, I can add it now though.

application/common/models/User.php
wcjr: yep, same response as above as well.

application/common/models/User.php
Baggerone: As long as the Broker sends the actual "expires" timestamp to SIL Auth, I think that should be OK.  The grace period here is just for the Broker to decide whether to block authentication, right?

wcjr:
@Baggerone ,  I can do it in this PR now if you and @fillup want it now.

application/common/models/User.php
wcjr: yes, we can ask @Baggerone .  The expiration date will be available in the returned `User`.

wcjr: the expiration will come out with the `User` as we discussed.  That will be enough, correct @Baggerone ?

Baggerone: Correct. (See my comment above.)

wcjr:
@Baggerone  @fillup , those configs are in now, take a look and let me know if we can merge.  Thanks

PR #61, opened 2017-05-05
uuid was needed after all, #60

### LINKED ISSUE #60: Bring UUID back
Check previous ramsey generator implementation for UUID.  Needs to come out of api as a `uuid` property.

wcjr:
resolves #60

### DIFF
--- a/application/common/models/User.php
+++ b/application/common/models/User.php
@@ -6,6 +6,7 @@ use Closure;
 use common\helpers\MySqlDateTime;
 use common\ldap\Ldap;
 use Exception;
+use Ramsey\Uuid\Uuid;
 use Yii;
 use yii\behaviors\AttributeBehavior;
 use yii\db\ActiveRecord;
@@ -93,10 +94,6 @@ class User extends UserBase
         }
     }
 
-    /**
-     * @param string $username
-     * @return User|null
-     */
     public static function findByUsername(string $username)
     {
         return User::findOne(['username' => $username]);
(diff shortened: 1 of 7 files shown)


application/common/models/User.php
```diff
@@ -93,10 +94,6 @@ protected function attemptPasswordMigration()
         }
     }
 
-    /**
```
forevermatt: Why the documentation removal?

application/features/user.feature
```diff
@@ -343,7 +344,7 @@ Feature: User
 #TODO: PUT with a change to an existing value on one of the unique fields should generate an error.
 #TODO: PUT with a change should update the last_changed data properly...as well as the field to be updated :-)
 
-
+#TODO: need to think about how to verify uuid comes back in calls.
```
forevermatt: Do you think it would also be worth (eventually) adding a test to verify that the uuid cannot be changed by being sent in (in all scenarios)?

application/common/models/User.php
wcjr: I assumed since the type and name of the param are already there in the function, the docs were pretty redundant.  Is that the case or do you needs those docs for something else?  

As for the `@return`, I assume that's very common pattern but if you think it's helpful for the consumer that may not know then I can add it back.

application/features/user.feature
wcjr: definitely, thanks:  https://github.com/silinternational/idp-id-broker/commit/e4145a6168fcf00e84babf9ec9ee5ce485053195

application/common/models/User.php
forevermatt: Hmm... I guess that is redundant, looking at the code more closely now. My IDE shows more helpful info about a function if it has the doc. comments, but I don't know whether that's worth the (very minor) DRY violation. Either way seems fine, in this case.

PR #62, opened 2017-05-08
response was not appropriate for an API in certain conditions.

wcjr:


wcjr:
@forevermatt , anything else you can think of needing in that response?  I'm assuming the response code of 500 and a message will be sufficient.

forevermatt:
@wcjr The common error data returned, for example, by ID Broker APIs is along these lines:
```
{
  "name": "Internal Server Error",
  "message": "Some error message",
  "code": 12345,
  "status": 500
}
```
It might be worth at least included `name` and `status` here. Thoughts?

PR #65, opened 2017-05-18
Move catch for EnvVarNotFoundException

forevermatt:
What do you two think about this slight refactor of catching the `EnvVarNotFoundException`s? Since calls from the console won't be HTTP requests, the http_response_code and JSON output formatting didn't quite seem applicable, so I moved the HTTP stuff to frontend/web/index.php, and put something in the console calls to return a non-zero exit code and a simple/clean error message.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 start: app
 
-app: db deps tables basemodels
+app: db deps
 	docker-compose up -d app
 
 deps:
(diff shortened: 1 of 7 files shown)


fillup:
Since this happens early in bootstrapping can we add a manual call to syslog to log the error so we can get better visibility into why it failed? We wont see the output directly in ECS so it'll be helpful to have it in logs.

I think we're talking about adding:

```php
openlog('id-broker', LOG_NDELAY | LOG_PERROR, LOG_USER);
syslog(LOG_CRIT, $e->getMessage());
closelog();

```

forevermatt:
@fillup I'll go ahead and try that now. Good idea.

PR #68, opened 2017-05-30
Decided we needed password metadata stored rather than derived.

wcjr:


application/common/models/User.php
```diff
@@ -196,44 +120,110 @@ private function validatePassword(): Closure
...
         return function ($attributeName) {
 
-            if ( ! $this->hasPasswordAlready()) {
+            if ($this->password_id === null) {
                 $this->attemptPasswordMigration();
             }
 
-            if (! password_verify($this->password, $this->password_hash)) {
+            /** @var Password $currentPassword */
+            $currentPassword = $this->getPassword()->one() ?? new Password();
```
forevermatt: You might want to add a `getCurrentPassword()` method to make sure (or maybe just make it clear) that the one you're getting is the current one, not one the possibly many of this `User`'s passwords stored in the `Password` table.

application/common/models/User.php
```diff
@@ -292,18 +282,9 @@ public function fields(): array
...
-            };
-
-            $fields['password_last_changed'] = function () {
-                /** @var $mostRecentPassword PasswordHistory */
-                $mostRecentPassword = $this->getPasswordHistories()
-                                           ->orderBy(['id' => SORT_DESC])
-                                           ->one();
-
-                return $mostRecentPassword->created_utc;
+        if ($this->password_id !== null) {
```
forevermatt: Out of curiosity, was there a reason to go back from `hasPasswordAlready()` to `$this->password_id !== null`? The function name seemed more readable to me, both for code internal to this class and for external code, but you often have good reasons for your changes, so I thought I'd ask.

application/common/models/User.php
```diff
@@ -313,20 +294,19 @@ public function fields(): array
...
+    private function updatePassword(): bool
     {
         $transaction = ActiveRecord::getDb()->beginTransaction();
 
         try {
-            $this->password_hash = password_hash($this->password, PASSWORD_DEFAULT);
 
-            if (! $this->saveHistory()) {
+            if (! $this->savePassword()) {
                 return false;
```
forevermatt: In this situation, do you need to end the transaction that was started just above? I know nothing seems to have been changed yet, but I didn't know if it affected the state of the database in some way that we would want to undo before moving on.

application/common/models/UserBase.php
```diff
@@ -40,13 +40,15 @@ public function rules()
...
+            [['password_id'], 'integer'],
             [['active', 'locked'], 'string'],
             [['last_changed_utc', 'last_synced_utc'], 'safe'],
             [['uuid'], 'string', 'max' => 64],
-            [['employee_id', 'first_name', 'last_name', 'display_name', 'username', 'email', 'password_hash'], 'string', 'max' => 255],
+            [['employee_id', 'first_name', 'last_name', 'display_name', 'username', 'email'], 'string', 'max' => 255],
             [['employee_id'], 'unique'],
             [['username'], 'unique'],
             [['email'], 'unique'],
+            [['password_id'], 'exist', 'skipOnError' => true, 'targetClass' => Password::className(), 'targetAttribute' => ['password_id' => 'id']],
```
forevermatt: I couldn't tell with only a moderate amount of digging: does this allow `password_id` to be `null` (as will happen for users that don't yet have a password in ID Broker)?

application/console/migrations/m170203_210542_create_initial_tables.php
```diff
@@ -39,38 +42,44 @@ private function createUserTable()
...
+    private function dropPasswordTable()
+    {
+        $this->dropTable('{{password}}');
     }
 
-    private function dropPasswordHistoryTable()
+    private function createForeignKeys()
     {
-        $this->dropForeignKey('fk_password_history_user_id','{{password_history}}');
+        $this->addForeignKey('fk_user_to_current_password','{{user}}','password_id','{{password}}','id','CASCADE','CASCADE');
```
forevermatt: Will we also want a foreign key from `password.user_id` to `user.id`? I don't actually know, since it's possible that having the foreign key going both directions will presumably affect the relationships Yii provides between those two models, and it might not properly infer them the way we have in mind.

forevermatt: P.S. If we do add that 2nd foreign-key relation, we'll probably want to be careful _not_ to use `'CASCADE'` for the ON DELETE argument to `addForeignKey(...)`... though I think using it for the ON UPDATE would be safe.

application/features/user.feature
```diff
@@ -29,7 +29,7 @@ Feature: User
         | locked       | no                    |
       And the following data is not returned:
         | property                |
-        | password_hash           |
+        | password_id             |
         | password_expires_at_utc |
```
forevermatt: How are we expecting the collection of password dates to come out of the API? Would it be in a `password` property?

api.raml
```diff
@@ -193,6 +186,30 @@ schemas:
             "required": false,
             "type": "string",
             "enum": ["yes", "no"]
+          },
+          "password": {
```
wcjr: yuk!  https://github.com/silinternational/idp-id-broker/commit/2b73555c3f2ae1d709b86a7d14fd8553761c482a 

We never put any conditional logic in place to suppress the password info on an update...is there some reason we'd want something like that?

application/common/models/PasswordBase.php
```diff
@@ -0,0 +1,64 @@
...
+            'created_utc' => Yii::t('app', 'Created Utc'),
+            'expiration_utc' => Yii::t('app', 'Expiration Utc'),
+            'grace_period_ends_utc' => Yii::t('app', 'Grace Period Ends Utc'),
+        ];
+    }
+
+    /**
+     * @return \yii\db\ActiveQuery
+     */
+    public function getUsers()
```
wcjr: Comes with other things that aren't ideal.  Not sure this is worth it afterall.  In the meantime, I went ahead and made some tweaks around the column naming for current password you requested.  https://github.com/silinternational/idp-id-broker/commit/c9752a3c1979dea7158a1eb89d0b6ca8742343c5

application/common/models/Password.php
```diff
@@ -0,0 +1,154 @@
...
+    public $password;
+
+    public function rules(): array
+    {
+        return ArrayHelper::merge([
+            [
+                'created_utc', 'default', 'value' => MySqlDateTime::now(),
+            ],
+            [
+                'expiration_utc', 'default', 'value' => MySqlDateTime::now(),
```
wcjr: yes, down in `behaviors()`.  Since they are required (from the base) they need to be populated to make it through validation.

application/common/models/User.php
wcjr: see https://github.com/silinternational/idp-id-broker/commit/c9752a3c1979dea7158a1eb89d0b6ca8742343c5 for related update

application/common/models/User.php
wcjr: I saw mixed uses, some calls to the method and some simply checking for `null`.  I hope with the new column name, this may not be as big an issue. (https://github.com/silinternational/idp-id-broker/commit/c9752a3c1979dea7158a1eb89d0b6ca8742343c5)

application/common/models/User.php
wcjr: you don't, not sure about the internal handling by the dbms for these but nothing gets written until the `commit()` runs.

application/common/models/UserBase.php
wcjr: yes

application/console/migrations/m170203_210542_create_initial_tables.php
wcjr: I don't think there's a use case for it @fillup but can certainly add it easy enough.  The only thing it might end up doing for us in the long run is frustrating us if we're not removing rows in the right order...

Agreed @forevermatt , if we do want it, we don't want users being removed when a password is removed :-)

application/features/user.feature
wcjr: yes, working out the tests on that stuff is going to have to wait since it may not be trivial.

api.raml
forevermatt: It may help us if we split out and/or clarify the purpose of some of the schemas in our RAML file. For example, `User` is being used both for what to supply when creating a User and for what comes back when getting a User... which I think won't be the same. (For example, AFAIK we won't want to let a POST provide a value for when the password expires... that's all calculated at the time the password is updated, and stored in the database then). At a subtler level, whether certain fields are required differs between those situations as well.

I'm not sure I actually address your question about suppressing the password info on an update... if not, could you help me understand what you mean?

application/common/models/PasswordBase.php
forevermatt: I like the cleanup/simplifications you made there in c9752a3  :-)

api.raml
wcjr: nevermind, I was not picking up on the fact that the schema was referring to what's coming in on the POST/PUT.  Corrected in https://github.com/silinternational/idp-id-broker/commit/c922f30c8971af1e3d0bd518d58a6e4e9be7817a

PR #74, opened 2017-06-12
Gives consumers the ability to search for users based on password expiration or newness criteria

### LINKED ISSUE #73: New API's needed
- (expired users) Retrieve all users whose accounts are active and current password is expiring on a specific date.
- (grace period has ended) Retrieve all users whose accounts are active and current password's grace period ends on a specific date.
- (new users) Retrieve all users whose whose accounts are active and oldest password was created on a specific date.

```
GET /user?active=yes&password.[created_utc|expiration_utc|grace_period_ends_utc]=YYYY-MM-DD
```

wcjr:
closes #73

### DIFF
--- a/application/common/helpers/MySqlDateTime.php
+++ b/application/common/helpers/MySqlDateTime.php
@@ -3,15 +3,21 @@ namespace common\helpers;
 
 class MySqlDateTime
 {
-    const MYSQL_FORMAT = 'Y-m-d H:i:s';
+    const MYSQL_DATETIME_FORMAT = 'Y-m-d H:i:s';
+    const MYSQL_DATE_FORMAT     = 'Y-m-d';
 
     public static function now()
     {
-        return gmdate(self::MYSQL_FORMAT);
+        return gmdate(self::MYSQL_DATETIME_FORMAT);
     }
 
-    public static function format(int $timestamp)
+    public static function today()
     {
-        return gmdate(self::MYSQL_FORMAT, $timestamp);
+        return gmdate(self::MYSQL_DATE_FORMAT);
+    }
+
+    public static function formatDate(int $timestamp)
+    {
+        return gmdate(self::MYSQL_DATE_FORMAT, $timestamp);
     }
 }
--- a/application/common/models/Password.php
+++ b/application/common/models/Password.php
@@ -133,9 +133,15 @@ class Password extends PasswordBase
     public function fields(): array
     {
         $fields = [
-            'created_utc',
-            'expires_on',
-            'grace_period_ends_on',
+            'created_utc' => function ($model) {
+                return "{$model->created_utc} UTC";
+            },
+            'expires_on' => function ($model) {
+                return "{$model->expires_on} 23:59:59 UTC";
+            },
+            'grace_period_ends_on' => function ($model) {
+                return "{$model->grace_period_ends_on} 23:59:59 UTC";
+            },
         ];
 
         return $fields;
--- a/application/common/models/Password.php
+++ b/application/common/models/Password.php
@@ -20,10 +20,10 @@ class Password extends PasswordBase
                 'created_utc', 'default', 'value' => MySqlDateTime::now(),
             ],
             [
-                'expiration_utc', 'default', 'value' => MySqlDateTime::now(),
+                'expires_on', 'default', 'value' => MySqlDateTime::today(),
             ],
             [
-                'grace_period_ends_utc', 'default', 'value' => MySqlDateTime::now(),
+                'grace_period_ends_on', 'default', 'value' => MySqlDateTime::today(),
             ],
             [
                 'password', 'required',
@@ -95,14 +95,14 @@ class Password extends PasswordBase
             'expirationTracker' => [
                 'class' => AttributeBehavior::className(),
                 'attributes' => [
-                    ActiveRecord::EVENT_BEFORE_INSERT => 'expiration_utc',
+                    ActiveRecord::EVENT_BEFORE_INSERT => 'expires_on',
                 ],
                 'value' => $this->expires()
             ],
             'gracePeriodTracker' => [
                 'class' => AttributeBehavior::className(),
                 'attributes' => [
-                    ActiveRecord::EVENT_BEFORE_INSERT => 'grace_period_ends_utc',
+                    ActiveRecord::EVENT_BEFORE_INSERT => 'grace_period_ends_on',
                 ],
                 'value' => $this->gracePeriodEnds()
             ],
@@ -114,7 +114,7 @@ class Password extends PasswordBase
         return function() {
             $lifespan = Yii::$app->params['passwordLifespan'];
 
-            return MySqlDateTime::format(strtotime($lifespan, strtotime($this->created_utc)));
+            return MySqlDateTime::formatDate(strtotime($lifespan, strtotime($this->created_utc)));
         };
     }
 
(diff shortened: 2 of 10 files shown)


application/common/helpers/MySqlDateTime.php
```diff
@@ -3,15 +3,21 @@
 
 class MySqlDateTime
 {
-    const MYSQL_FORMAT = 'Y-m-d H:i:s';
+    const MYSQL_DATETIME_FORMAT = 'Y-m-d H:i:s';
+    const MYSQL_DATE_FORMAT     = 'Y-m-d';
```
forevermatt: @fillup I'm a little concerned about just using a date (not a specific point in time) for password expirations. Depending on how `expires_on` is interpreted (beginning of day vs. end of day) and what the server's timezone is set to (which affects how `strtotime(...)` is going to interpret these dates), we could have a nearly 2-day window between when we meant for the password to expire and when some other code using that value may actually consider it to have expired. If a potentially 2-day margin-of-error on password expirations is acceptable (and it may be, given that our servers typically default to UTC, so the error is more likely just an off-by-one margin-of-error), then I'm fine with this. If we want to reduce/remove that range of possible interpretations of the expiration date, it may be worth specifying it as a full date/time string (with timezone, ideally), such as `2018-05-24 00:00:00 UTC` (or `2018-05-24 23:59:59 UTC`, if you want end-of-day), and then just showing the relevant information to the end user.

application/common/models/Password.php
```diff
@@ -114,7 +114,7 @@ private function expires(): Closure
         return function() {
             $lifespan = Yii::$app->params['passwordLifespan'];
 
-            return MySqlDateTime::format(strtotime($lifespan, strtotime($this->created_utc)));
+            return MySqlDateTime::formatDate(strtotime($lifespan, strtotime($this->created_utc)));
```
forevermatt: If PHP's default timezone is set to something other than UTC, this may not give you back the point-in-time that you expect. Giving `strtotime(...)` a string that doesn't specify a timezone results in it assuming the given time is in PHP's default timezone. I think we have that set to UTC for our PHP apps, so this may be fine.

application/common/models/User.php
```diff
@@ -210,7 +210,7 @@ private function validateExpiration(): Closure
     {
         return function ($attributeName) {
             if ($this->currentPassword !== null) {
-                $gracePeriodEnds = strtotime($this->currentPassword->grace_period_ends_utc);
+                $gracePeriodEnds = strtotime("{$this->currentPassword->grace_period_ends_on} 23:59:59");
```
forevermatt: If we make `... 23:59:59` instead be `... 23:59:59 UTC`, it might further reduce timezone offset errors.

application/common/models/User.php
```diff
@@ -343,6 +343,55 @@ private function savePassword()
         return true;
     }
 
+    public static function getExpiringUsers($criteria): array
```
forevermatt: Is the goal of `getExpiringUsers(...)` to get users that will expire within a day? The desired timeframe seemed ambiguous to me, so I wasn't sure.

application/frontend/controllers/UserController.php
```diff
@@ -81,4 +84,16 @@ public function actionUpdatePassword(string $employeeId)
 
         return $user;
     }
+
+    public function actionExpiring(): array
```
forevermatt: You'll probably want to add these `/user/expiring` and `/user/new` endpoints to the `api.raml` file.

application/common/models/User.php
```diff
@@ -343,6 +343,55 @@ private function savePassword()
...
+                default:
+                    // if no criteria names match, this will ensure an empty result is returned
+                    $users->where('0=1');
+            }
+        }
+
+        return $users->all();
+    }
+
+    public static function getNewUsers($createdOn): array
```
forevermatt: This looks like it will return the users whose first password was created on the specified date (not necessarily users that we added on the specified date). Was that the intent? Since a user record might not have a password until long after the user record is added to ID Broker (such as if the user doesn't log in for quite some time), it seemed like the name and the actual effect didn't quite line up for this function.

application/common/models/User.php
wcjr: https://github.com/silinternational/idp-id-broker/commit/4ee36998709846a5e0da1b6f3c297340e5e4f7d3

application/common/models/User.php
wcjr: That is the intent for now until we find a need to include the idea of newly created users in our definition of "new"

application/common/helpers/MySqlDateTime.php
fillup: @wcjr and I talked about that in the DB we would store only date, but we would return from the api the date plus 23:59:59 UTC so from a display standpoint we could be clear about exactly when it expires. I think Billy forgot about that and I'm not sure it's worth changing again since it is just hard coded information anyway. From an end user view we can add that it expires just before midnight UTC.

application/common/models/Password.php
fillup: Our base image is set to UTC so I think it should be fine.

application/common/models/Password.php
```diff
@@ -20,10 +20,10 @@ public function rules(): array
                 'created_utc', 'default', 'value' => MySqlDateTime::now(),
             ],
             [
-                'expiration_utc', 'default', 'value' => MySqlDateTime::now(),
```
fillup: Is this right the default expiration and grace period ending is "today"?

application/common/models/User.php
fillup: We're matching for a specific date, but including the time for clarity to the end user. This is what I expected everywhere for returned expiration and grace times actually.

application/common/models/User.php
fillup: @forevermatt that is the desired behavior. The "new" concept is intended for triggering a welcome email and we want that to happen as close to when the user first became aware of their account and used it. We don't want to trigger emails when they are first created in the DB since that could be at least weeks before it is announced to users and an automated email would look like a phishing attack. We want to keep the context of creating a password and receiving a welcome email close together.

application/common/helpers/MySqlDateTime.php
wcjr: very possible, changes in here:  https://github.com/silinternational/idp-id-broker/commit/0a55655953b0098a7f5ec423c5835e2461e10f41

application/common/models/Password.php
wcjr: in that case, I won't tweak that to ensure UTC at this time.

application/common/models/User.php
wcjr: This sounds like a question of clarity around what/how `$criteria` is to be used.  Docs should help with that, see https://github.com/silinternational/idp-id-broker/commit/c79866937c503de1b5d60cb4a75d358f146cec5b

application/common/models/Password.php
wcjr: it's just a default value, it will be dealt with appropriately in the behaviors before insert.

application/common/models/User.php
wcjr: let's try a different endpoint:  https://github.com/silinternational/idp-id-broker/commit/293ff2edaf4d80a6b528851090384f097217deb1

application/frontend/controllers/UserController.php
wcjr: https://github.com/silinternational/idp-id-broker/commit/d31e4b43ce1b1ec50cd88e495b7869ec139ec115

PR #81, opened 2017-07-17
Fix Makefile and docker-compose.yml to have consistent LDAP expectations

forevermatt:


forevermatt:
Okay, @wcjr, this is probably the better solution, as you said. How's it look now?

PR #87, opened 2017-08-14
Send (via Email Service) and log emails

forevermatt:


forevermatt:
Note that this does **not** have actual wording in the welcome or invite emails... just view files where such wording will eventually go. There are `@todo` items in those view files, to help us remember that.

application/common/components/Emailer.php
```diff
@@ -24,7 +60,7 @@ class Emailer extends Component
      * @param string $subject The subject.
      * @param string $htmlBody The email body (as an HTML string).
      */
-    public function email(string $toAddress, string $subject, string $htmlBody)
+    protected function email(string $toAddress, string $subject, string $htmlBody)
```
fillup: is there a reason we're only supporting html emails? the mailer service supports html and text and I assumed we'd render both versions

application/common/models/User.php
```diff
@@ -212,6 +223,38 @@ public static function findByEmail(string $email)
         return User::findOne(['email' => $email]);
     }
 
+    /**
+     * Get the list of attributes about this User that are safe to include in
+     * an email to them.
+     *
+     * NOTE: The resulting array uses camelCased attribute names for keys, so
+     *       the User's "employee_id" with have a key of "employeeId".
```
Baggerone: "will have" ?

application/common/components/Emailer.php
forevermatt: Since this class is purely for use from within ID Broker, and it seemed likely that we will always be providing HTML email bodies, I didn't see a reason to allow us (from within ID Broker) to specify both an HTML body and a text body.

I did wonder about whether the Email Service would automatically generate a text body from the HTML body (if no text body was provided) or whether I should do that here. Would you like me to change it to something like the following?
```php
$this->getEmailServiceClient()->email([
    'to_address' => $toAddress,
    'subject' => $subject,
    'html_body' => $htmlBody,
    'text_body' => strip_tags($htmlBody),
]);
```

forevermatt: The [strip_tags](http://php.net/manual/en/function.strip-tags.php) function isn't perfect for converting HTML to text, but for our emails it will probably cover 98%+ of what we'll need to do (missing things like `&quot;` and other HTML entities).

forevermatt:
@fillup How does 6aa4e54 look?

PR #88, opened 2017-08-16
Use email-service log target

forevermatt:


application/common/config/main.php
```diff
@@ -2,8 +2,9 @@
 
 use common\components\Emailer;
 use common\ldap\Ldap;
-use Sil\JsonSyslog\JsonSyslogTarget;
+use Sil\JsonLog\target\JsonSyslogTarget;
 use Sil\Log\EmailTarget;
```
wcjr: is `EmailTarget` still used?

application/common/config/main.php
```diff
@@ -95,15 +96,39 @@
                     },
                 ],
                 [
-                    'class' => EmailTarget::class,
-                    'categories' => ['application'], // stick to messages from this app, not all of Yii's built-in messaging.
-                    'logVars' => [], // no need for default stuff: http://www.yiiframework.com/doc-2.0/yii-log-target.html#$logVars-detail
+                    'class' => EmailServiceTarget::class,
```
wcjr: is removing `categories` here intentional?

application/common/config/main.php
```diff
@@ -95,15 +96,39 @@
...
+                        'yii\web\HttpException:409',
+                        'Sil\EmailService\Client\EmailServiceClientException',
+                    ],
                     'levels' => ['error'],
+                    'logVars' => [], // Disable logging of _SERVER, _POST, etc.
                     'message' => [
-                        'from' => $mailerUsername,
                         'to' => $notificationEmail,
-                        'subject' => "ERROR - $idpName-id-broker [".YII_ENV."] Error",
+                        'subject' => 'ALERT - ' . $idpName . ' ID Broker [' . YII_ENV .']',
```
wcjr: why change to `ALERT` even though the level is still `'error'`?

application/common/config/main.php
forevermatt: Nope, good catch. I'll fix that.

application/common/config/main.php
forevermatt: No, it wasn't. I simply copied/adapted this `EmailServiceTarget` configuration from [idp-pw-api's version](https://github.com/silinternational/idp-pw-api/blob/develop/application/common/config/main.php#L102-L144)... though it's interesting that that repo didn't limit by category. I'll add that back here.

application/common/config/main.php
forevermatt: Again, this was just an accidental carry-over from idp-pw-api's version. I'll change it back to "ERROR" here.

forevermatt:
@wcjr - I've made updates per your feedback. Would you like to re-review this PR?

wcjr:
Nope, go for it

PR #96, opened 2017-11-08
MFA rate limit

forevermatt:


application/common/components/Emailer.php
```diff
@@ -13,6 +13,7 @@
 class Emailer extends Component
 {
     const SUBJECT_INVITE_DEFAULT = 'Your New Account';
+    const SUBJECT_MFA_RATE_LIMIT_DEFAULT = 'Incorrect 2-Step Verification Attempts';
```
fillup: Does this make sense? Should be be `Too many 2-Step Verification Attempts`?

PR #101, opened 2017-11-30
increasing mfa related logging for succes events

fillup:


application/common/models/Mfa.php
```diff
@@ -145,6 +171,13 @@ public function verify($value): bool
                 ]);
             }
             $this->clearFailedAttempts('after successful verification');
+
+            \Yii::warning([
+                'action' => 'verify mfa',
+                'type' => $this->type,
+                'user' => $this->user->email,
+                'status' => 'success',
```
forevermatt: Our MFA module (in SSP) is already logging MFA successes and failures, though it doesn't hurt for Broker to log them as well. In that case, do you also want broker to log MFA failures (just before line 185), not just successes?

fillup: yeah good catch, I'll add that too

PR #102, opened 2017-11-30
Feature/welcome email

fillup:


application/common/mail/welcome.html.php
```diff
@@ -0,0 +1,86 @@
...
+ */
+?>
+<p>
+    Dear <?=yHtml::encode($displayName)?>,
+</p>
+<p>
+    Congratulations! You have logged into your new <?=yHtml::encode($idpDisplayName)?> account for the first time.
+    <?=yHtml::encode($idpDisplayName)?> is in the process of transitioning from logging into websites using
+    an Insite account to this new "<?=yHtml::encode($idpDisplayName)?> account". Starting in January 2018 the option
+    to login to websites using your Insite account will go away. This new <?=yHtml::encode($idpDisplayName)?>
```
forevermatt: Minor tweak: `to login to websites` --> `to log in to websites` (since we've been making that distinction in other communications).

application/common/mail/welcome.html.php
```diff
@@ -0,0 +1,86 @@
...
+    <li>Enter the email address or phone number you wish to use and click "Send Code"</li>
+    <li>If you entered an email address, check the inbox for that email address for a new email and retrieve the code
+        from that email.</li>
+    <li>If you entered a phone number, watch for a text message or phone call and take note of the code you receive.</li>
+    <li>Enter the verification code into the form on your screen and click "Verify".</li>
+
+</ol>
+
+<p>
+    <strong>Enable 2-Step Verification for enhanced security</strong>
```
forevermatt: Should we somehow indicate that this is not a required step, merely an recommended step?

application/common/mail/welcome.text.php
```diff
@@ -0,0 +1,74 @@
...
+ * @var string $supportName
+ * @var bool   $isMfaEnabled
+ */
+?>
+Dear <?=yHtml::encode($displayName)?>,
+
+Congratulations! You have logged into your new <?=yHtml::encode($idpDisplayName)?> account for the first time.
+<?=yHtml::encode($idpDisplayName)?> is in the process of transitioning from logging into websites using
+an Insite account to this new "<?=yHtml::encode($idpDisplayName)?> account". Starting in January 2018 the option
+to login to websites using your Insite account will go away. This new <?=yHtml::encode($idpDisplayName)?>
```
forevermatt: Same as above (`login` to `log in`), if you decide to bother with that change.

application/common/models/User.php
```diff
@@ -580,7 +584,7 @@ public function attributeLabels()
     public function isPromptForMfa(): bool
     {
         if ($this->scenario == self::SCENARIO_AUTHENTICATE) {
-            if (strtolower($this->require_mfa == 'yes') || count($this->mfas) > 0) {
+            if ($this->require_mfa == 'yes' || count($this->getVerifiedMfaOptions()) > 0) {
```
forevermatt: This `==` could probably be a `===`, unless you know of some reason not to.

application/common/components/Emailer.php
```diff
@@ -260,7 +265,25 @@ public function shouldSendInviteMessageTo($user, $isNewUser)
     public function shouldSendPasswordChangedMessageTo($user, $changedAttributes)
```
Baggerone: Sorry - this wasn't a PR change, but I was wondering why $user is still a parameter if it's not being used in the function.
That applies to the next function also.

application/common/mail/welcome.html.php
```diff
@@ -0,0 +1,86 @@
...
+ * @var string $supportEmail
+ * @var string $supportName
+ * @var bool   $isMfaEnabled
+ */
+?>
+<p>
+    Dear <?=yHtml::encode($displayName)?>,
+</p>
+<p>
+    Congratulations! You have logged into your new <?=yHtml::encode($idpDisplayName)?> account for the first time.
```
Baggerone: Would it be good to start with ...
`$idpEncodedName = yHtml::encode($idpDisplayName);`  
and then use that throughout the message. (The same goes for the text email message below)

Baggerone: Instead of "account" I wonder if we could come up with a clearer wording, e.g. "online account" or "web account". (Applies to the text email message too.)

application/common/mail/welcome.html.php
```diff
@@ -0,0 +1,86 @@
...
+    <?=yHtml::encode($idpDisplayName)?> is in the process of transitioning from logging into websites using
+    an Insite account to this new "<?=yHtml::encode($idpDisplayName)?> account". Starting in January 2018 the option
+    to login to websites using your Insite account will go away. This new <?=yHtml::encode($idpDisplayName)?>
+    account is not the same as your email account or computer account, it is a new corporate identity for use with
```
Baggerone: Remove "use with"? (Applies to the text email message too.)

application/common/components/Emailer.php
forevermatt: Ohhh... good catch. That was partially a copy-paste oversight (for this function), but a logic oversight for the next function. It should also be checking whether the user has already received a welcome message.

forevermatt: On further thought, since there is no supported way for a user to stop having a password (which is what it would take to allow them to later re-receive the welcome email), the extra code for preventing sending a welcome message to someone who has already received one seems unnecessary. If a user were somehow deleted from the system (not currently supported), and then were later re-added, re-sending them the welcome message doesn't seem like a bad thing.

PR #110, opened 2017-12-06
Feature/google analytics

Baggerone:


application/console/controllers/CronController.php
```diff
@@ -14,4 +18,61 @@ public function actionRemoveOldUnverifiedRecords()
...
+     *  - active users that have a verified Mfa of any type
+     *  - active users with a backup code Mfa
+     *  - active users with a verified totp Mfa
+     *  - active users with a verified u2f Mfa
+     *
+     * @throws \Exception
+     */
+    public function actionGoogleAnalytics()
+    {
+        $trackingId = Env::get('GA_TRACKING_ID', null); // 'UA-12345678-12'
```
fillup: Shouldn't this be in `common/config/main.php` as part of `params`, perhaps a `googleAnalytics` element with `trackingId` and `clientId` children? I believe all our other `Env` usage is in the config for central view and usage.

application/console/controllers/CronController.php
```diff
@@ -27,18 +26,30 @@ public function actionRemoveOldUnverifiedRecords()
...
+     *            'results' => $response->getDebugResponse(),
+     *     ]);
+     *
+     *
      * @throws \Exception
      */
     public function actionGoogleAnalytics()
     {
-        $trackingId = Env::get('GA_TRACKING_ID', null); // 'UA-12345678-12'
+        $trackingId = \Yii::$app->params('gaTrackingId'); // 'UA-12345678-12'
```
forevermatt: That's interesting... I didn't know `params` worked as a function call (`params('something')`). I generally treat it as an array (`params['something']`).

application/console/config/main.php
```diff
@@ -1,4 +1,13 @@
 <?php
+
+
+use Sil\PhpEnv\Env;
+
+/* For pushing data to Google Analytics */
+$gaTrackingId = Env::get('GA_TRACKING_ID', null); // 'UA-12345678-12'
+$gaClientId = Env::get('GA_CLIENT_ID', null); // 'IDP_ID_BROKER_LOCALHOST'
```
forevermatt: The example values in your comments may be worth moving to the `local.env.dist` next to each of these optional env. vars.

PR #111, opened 2017-12-07
Added functions and tests for more Google analytics events

Baggerone:
Users with password.
Users with require_mfa.
Average Mfas per User with Mfas.

### DIFF
--- a/application/features/analytics.feature
+++ b/application/features/analytics.feature
@@ -62,3 +62,40 @@ Feature: User
       And I create a new user
     When I get the count of active users with a verified u2f mfa
     Then the count of active users with a verified u2f mfa should be 1
+
+  Scenario: Get a count of active users with a password
+    Given that no mfas or users exist
+      And I create a new user with a password
+      And I create a new user with a password
+      And I create a new user without a password
+      And I create a new user without a password
+    When I get the count of active users with a password
+    Then the count of active users with a password should be 2
+
+  Scenario: Get the average number of mfas per active user with a verified mfa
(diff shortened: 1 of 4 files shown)


application/features/analytics.feature
```diff
@@ -61,4 +61,41 @@ Feature: User
...
-    Then the count of active users with a verified u2f mfa should be 1
\ No newline at end of file
+    Then the count of active users with a verified u2f mfa should be 1
+
+  Scenario: Get a count of active users with a password
+    Given that no mfas or users exist
+      And I create a new user with a password
+      And I create a new user with a password
+      And I create a new user without a password
+      And I create a new user without a password
```
forevermatt: It's probably not a big deal, but it might be worth having a different number of users with and without a password in case we ever update that logic and get something backwards (like our recent `'yes'` acting like `0` scenario), just to help ensure this test would catch that.

application/features/analytics.feature
```diff
@@ -61,4 +61,41 @@ Feature: User
...
+      And I create a new user
+    When I get the average number of mfas per active user with mfas
+    Then the average number of mfas per active user with mfas should be 2
+
+  Scenario: Get a count of active users with require mfa
+    Given that no mfas or users exist
+      And I create a new user with require mfa
+      And I create a new user with require mfa
+      And I create a new user
+      And I create a new user
```
forevermatt: As above, it may be worth having a different number of each type of user, to make sure the logic is counting the correct set of users.

PR #112, opened 2017-12-08
Release 3.3.0 (Google Analytics and Log stack traces)

Baggerone:


application/console/config/main.php
```diff
@@ -1,4 +1,10 @@
 <?php
+
+
+use Sil\PhpEnv\Env;
```
forevermatt: I don't think we need this `use` statement anymore, since the env stuff was moved to the common config file.

dockerbuild/broker-cron
```diff
@@ -1,3 +1,4 @@
 # m h dom mon dow user	command
 10 */6 * * * root /data/yii cron/remove-old-unverified-records 2>&1 | /usr/bin/logger -t id-broker-cron
 #* * * * * root /data/yii cron/remove-old-unverified-records 2>&1 | /usr/bin/logger -t id-broker-cron
+* 1 * * * root /data/yii cron/google-analytics 2>&1 | /usr/bin/logger -t id-broker-cron
```
forevermatt: Looks like this should be `0 1 * * *`

PR #114, opened 2017-12-13
Small email cleanups

forevermatt:


application/common/mail/invite.html.php
```diff
@@ -26,7 +26,18 @@
 ?>
 <p>Dear <?= yHtml::encode($displayName) ?>,</p>
 
-<p>A new <?= yHtml::encode($idpDisplayName) ?> account has been created for you and is ready for you to create your password. After creating your password you'll also be given the opportunity to set up recovery methods in case you ever forget your password. We highly recommended that you set up at least one or two recovery methods. You can also enhance the security of your account by enabling 2-Step Verification which will help ensure bad guys cannot get into your account even if they guess your password.</p>
+<p>
+    A new <?= yHtml::encode($idpDisplayName) ?> Identity account has been created for you and is ready for you to set
+    your password. Please note this account is not the same as your <?= yHtml::encode($idpDisplayName) ?> email account and therefore
```
forevermatt: Could we change...
`Please note this account is not the same as your <?= yHtml::encode($idpDisplayName) ?> email account and therefore passwords are not kept in sync. Instead...`
... to simply be the following?
`Please note that this is not an email account. Instead...`

PR #115, opened 2017-12-15
Feature/more email types

Baggerone:


application/common/components/Emailer.php
```diff
@@ -19,6 +21,21 @@ class Emailer extends Component
...
+    const SUBJECT_GET_BACKUP_CODES_DEFAULT = 'Get printable codes for your %s account';
+    const SUBJECT_REFRESH_BACKUP_CODES_DEFAULT = 'Get a new set of printable codes for your %s account';
+    const SUBJECT_LOST_SECURITY_KEY_DEFAULT = 'Have you lost the security key you use with your %s account';
+
+    const SUBJECT_MFA_OPTION_ADDED_DEFAULT = 'A 2-Step Verification option was added to your %s account';
+    const SUBJECT_MFA_OPTION_REMOVED_DEFAULT = 'A 2-Step Verification option was removed from your %s account';
+    const SUBJECT_MFA_ENABLED_DEFAULT = '2-Step Verification was enabled on your %s account';
+    const SUBJECT_MFA_DISABLED_DEFAULT = '2-Step Verification was disabled on your %s account';
+
+    /* The number of days of not using a security key after which we email the user */
```
forevermatt: It just occurred to me that, if someone primarily uses their security key (U2F) and has their browser remember them for 30 days, then if they use some other computer (like the one in LSC 108) and use a code because it's more convenient, they'll probably get this email.

We may want to bump this up from 14 days to something noticeably over 30 days. Perhaps 40?

application/common/components/Emailer.php
```diff
@@ -285,6 +340,134 @@ public function shouldSendWelcomeMessageTo($user, $changedAttributes)
             && array_key_exists('current_password_id', $changedAttributes)
             && is_null($changedAttributes['current_password_id']);
     }
+
+    /**
+     * @param User $user The User in question.
+     * @return bool
+     */
+    public function shouldSendGetBackupCodesMessageTo($user)
```
forevermatt: We may need to add something like...
```
...
&& !$user->hasReceivedMessage(EmailLog::MESSAGE_TYPE_GET_BACKUP_CODES)
```
... to each of these new `shouldSend...()` methods so that we don't send these emails every night (or whatever cron schedule we have these on) as soon as the user matches the other criteria.

We may actually need to add a `hasReceivedMessageRecently(string $messageType, int $recentDays)` method so that we can send these reminder emails once every X days.

application/common/models/Mfa.php
```diff
@@ -279,6 +285,10 @@ public static function create(int $userId, string $type, string $label = null):
             'status' => 'success',
         ]);
 
+
+        /* @todo uncomment this line when the emails are ready */
+        // self::sendAppropriateMessages($user, self::EVENT_TYPE_CREATE);
```
forevermatt: This seems like it would be better placed in an `afterSave()` method on this `Mfa` model (which does include information about whether it was an insert or an update). That way, we know that this `sendAppropriateMessages(...)` will be called for each new Mfa, even if it was created outside of this static `Mfa::create(...)` method.

application/common/mail/get-backup-codes.html.php
```diff
@@ -0,0 +1,10 @@
+<?php
+use yii\helpers\Html as yHtml;
+
+/**
+ * @var string $displayName
```
forevermatt: If it's helpful, you can copy the list of variables from one of the other emails that is already written. Every email template will get that same list of variables.

application/common/models/User.php
```diff
@@ -414,6 +414,32 @@ public function getVerifiedMfaOptions()
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /*
+     * @return int the count of a user's Mfa backup codes
+     */
+    public function countMfaBackupCodes()
```
forevermatt: We were already showing this number on the PW Manager profile page. Do we need to update code somewhere else in ID Broker to also use this function now (so that the logic stays the same for both)?

application/console/controllers/CronController.php
```diff
@@ -82,4 +83,27 @@ public function actionGoogleAnalytics()
...
+
+
+    public function actionSendDelayedMfaRelatedEmails()
+    {
+        /* @var $emailer Emailer */
+        $emailer = \Yii::$app->emailer;
+
+        $users = User::findAll();
+
+        foreach ($users as $user) {
```
forevermatt: Given that this could be processing a list of thousands of users, there may be a way to get the list of Users as some kind of iterator, which only retrieves them as you're ready for them (or perhaps in batches, I'm not sure). While that may slow the process down, it would avoid having to instantiate a User model for every user in the database all at once, which I suspect will be the more likely limiting factor (RAM vs. CPU).

application/console/migrations/m171211_130001_add_several_mfa_email_log_types.php
```diff
@@ -0,0 +1,30 @@
...
+class m171211_130001_add_several_mfa_email_log_types extends Migration
+{
+
+    public function safeUp()
+    {
+        $this->alterColumn(
+            '{{email_log}}',
+            'message_type',
+            "enum('invite','welcome','mfa-rate-limit','password-changed','get-backup-codes','refresh-backup-codes'," .
+                "'lost-security-key','mfa-required','mfa-option-added','mfa-option-removed'," .
```
forevermatt: Was `mfa-required` supposed to be in this list? I don't see any other uses of that email-log type.

application/common/components/Emailer.php
Baggerone: @fillup Input on this?

application/features/email.feature
```diff
@@ -65,3 +65,154 @@ Feature: Email
...
+      And there are <backupCodes> backup codes
+    When I check if a refresh backup codes email should be sent
+    Then I see that a refresh backup codes email <shouldOrNot> be sent
+
+    Examples:
+      | sendRefreshBackupCodesEml | backupExistsOrNot | backupCodes | shouldOrNot |
+      | to send                   | does NOT exist    | 0           | should NOT  |
+      | to send                   | does NOT exist    | 0           | should NOT  |
+      | to send                   | does exist        | 3           | should      |
+      | to send                   | does exist        | 4           | should NOT  |
```
forevermatt: Based on the comment by [`MINIMUM_BACKUP_CODES_BEFORE_NAG`](https://github.com/silinternational/idp-id-broker/pull/115/files#diff-15c4448209de48489ad86a814d4a4dbcR36), it sounds like we should send the refresh-backup-codes email when the user has 4 backup codes left, but the tests say to send when they have 3 codes left. Was that a typo in the test, or is there a better name for `MINIMUM_BACKUP_CODES_BEFORE_NAG` to make it clear whether it means `=` or `<=`?

application/features/email.feature
```diff
@@ -65,3 +65,154 @@ Feature: Email
...
+      | NOT to send               | does exist        | 2           | should NOT  |
+
+  Scenario Outline: When to send mfa option added emails
```
forevermatt: These tests related to sending mfa enabled/disabled/added/removed emails seem to expect these emails to be sent under different circumstances than I expected. For example, this test seems to indicate that an mfa-option-added email should only be sent when there were already at least 2 MFA options in place and another is added, rather than when there is at lease 1 MFA option in place and another is added.

Could these tests be simplified to only care about (A) how many verified MFA options were already present and (B) whether we are adding or deleting an MFA option? (Feel free to suggest a different approach, or to help me better understand the current approach.)

On a related note, should we be sending these emails when we add the MFA option, or when we verify it?

application/common/components/Emailer.php
Baggerone: I made this a config

application/common/components/Emailer.php
```diff
@@ -19,6 +21,21 @@ class Emailer extends Component
...
+    const SUBJECT_REFRESH_BACKUP_CODES_DEFAULT = 'Get a new set of printable codes for your %s account';
+    const SUBJECT_LOST_SECURITY_KEY_DEFAULT = 'Have you lost the security key you use with your %s account';
+
+    const SUBJECT_MFA_OPTION_ADDED_DEFAULT = 'A 2-Step Verification option was added to your %s account';
+    const SUBJECT_MFA_OPTION_REMOVED_DEFAULT = 'A 2-Step Verification option was removed from your %s account';
+    const SUBJECT_MFA_ENABLED_DEFAULT = '2-Step Verification was enabled on your %s account';
+    const SUBJECT_MFA_DISABLED_DEFAULT = '2-Step Verification was disabled on your %s account';
+
+    /* The number of days of not using a security key after which we email the user */
+    const LOST_SECURITY_KEY_EMAIL_DAYS = 14;
```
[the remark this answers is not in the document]
Baggerone: I made these configs and public properties

application/common/helpers/MySqlDateTime.php
```diff
@@ -60,4 +60,19 @@ public static function formatDateForHumans(string $timestamp = null)
...
+     * @param string $dbDate formated datetime from database
+     * @param int $recentDays
+     * @return bool
+     */
+    public static function dateIsRecent(string $dbDate, int $recentDays)
+    {
+        $dtInterval = '-' . $recentDays . ' days';
+        $recentDate = self::relative($dtInterval);
+
+        return $dbDate >= $recentDate;
```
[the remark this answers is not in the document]
Baggerone: Done

application/common/mail/get-backup-codes.html.php
Baggerone: Done

application/common/models/Mfa.php
Baggerone: Done

application/common/models/User.php
Baggerone: Leaving as is, after discussing it.

application/console/controllers/CronController.php
Baggerone: Switched to use $query->each()

application/console/migrations/m171211_130001_add_several_mfa_email_log_types.php
Baggerone: Removed.  (There had been another type of email to send. But, it probably won't work here.)

application/features/bootstrap/EmailContext.php
```diff
@@ -253,4 +379,339 @@ public function weAreConfiguredNotToSendWelcomeEmails()
...
+    }
+
+    /**
+     * @Given there are :arg1 backup codes
+     */
+    public function thereAreXBackupCodes($count)
+    {
+        $backupMfa = $this->getMfa(Mfa::TYPE_BACKUPCODE);
+
+        if (empty($backupMfa)) {
```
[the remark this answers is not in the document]
Baggerone: Fixed to throw an exception if there is no MFA backup option but some codes are requested.

application/features/email.feature
```diff
@@ -65,3 +65,154 @@ Feature: Email
...
+      And a u2f mfa option <u2fExistsOrNot>
+      And a backup code mfa option was used <backupUsedDaysAgo> days ago
+      And a totp mfa option was used <totpUsedDaysAgo> days ago
+    When I check if a lost security key email should be sent
+    Then I see that a lost security key email <shouldOrNot> be sent
+
+  Examples:
+      | sendLostKeyEml | u2fExistsOrNot | backupUsedDaysAgo | totpUsedDaysAgo | shouldOrNot |
+      | to send        | does NOT exist | 2                 | 3               | should NOT  |
+      | to send        | does NOT exist | 2                 | 3               | should NOT  |
```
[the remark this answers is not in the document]
Baggerone: Fixed

application/features/email.feature
Baggerone: Clarified the comment (i.e. 4 is OK, 3 triggers a nag)

application/features/email.feature
forevermatt: Here's an example of what I was thinking:
```
  Scenario Outline: When to send mfa option added emails
    Given we are configured <sendMfaOptionAddedEml> mfa option added emails
      And a user already has <numMfasAlready> verified mfa options
      And they <addOrRemove> an mfa option
    When I check if a mfa option added email should be sent
    Then I see that a mfa option added email <shouldOrNot> be sent

    Examples:
      | sendMfaOptionAddedEml | numMfasAlready | addOrRemove | shouldOrNot |
      | to send               | 0              | add         | should NOT  |
      | to send               | 1              | add         | should      |
      | to send               | 2              | add         | should      |
      | to send               | 0              | remove      | should NOT  |
      | to send               | 1              | remove      | should NOT  |
      | to send               | 2              | remove      | should NOT  |
      | NOT to send           | 0              | add         | should NOT  |
      | NOT to send           | 1              | add         | should NOT  |
      | NOT to send           | 2              | add         | should NOT  |
      | NOT to send           | 0              | remove      | should NOT  |
      | NOT to send           | 1              | remove      | should NOT  |
      | NOT to send           | 2              | remove      | should NOT  |
```
You could then even use this same set of examples for all four of these types of emails (mfa added, mfa enabled, mfa removed, mfa disabled) and only have to change the `shouldOrNot` column's values (and the `sendMfaOptionAddedEml` column's title).

Baggerone: For now, I'm just going to reword the Scenario and the "And the mfa even type ..." line to clarify that it already happened in the past.

application/common/models/Mfa.php
```diff
@@ -400,4 +415,25 @@ public static function removeOldUnverifiedRecords($maxAgeHours = 2)
             'count' => $numDeleted,
         ]);
     }
+
+
+    protected function sendAppropriateMessages($user, $eventType)
```
forevermatt: Both places we call this, we're using `self::sendAppropriateMessages(...)`, so it looks like it needs to be changed from `protected function` to `protected static function`.

application/features/bootstrap/EmailContext.php
```diff
@@ -228,6 +231,52 @@ public function thatUserDoesNotHaveAPassword()
...
+            'message_type' => $messageType,
+        ]);
+
+        if ($daysAgo > 0) {
+            $diffConfig = "-" . $daysAgo . " days";
+            $dbDate = MySqlDateTime::relative($diffConfig);
+
+            $emailLog->sent_utc = $dbDate;
+        }
+        echo PHP_EOL . "Date: " . $emailLog->sent_utc;
```
forevermatt: Did you mean to leave this `echo` in here?

application/features/bootstrap/EmailContext.php
```diff
@@ -228,6 +231,52 @@ public function thatUserDoesNotHaveAPassword()
...
+     */
+    public function anEmailHasBeenSentToThatUser(string $messageType)
+    {
+
+        $emailLog = new EmailLog([
+            'user_id' => $this->tempUser->id,
+            'message_type' => $messageType,
+        ]);
+
+        $emailLog->save();
```
forevermatt: It might merit wrapping this in an `Assert::true(...)` to make sure the email log is successfully saved.

application/features/bootstrap/EmailContext.php
```diff
@@ -228,6 +231,52 @@ public function thatUserDoesNotHaveAPassword()
...
+    }
+
+    /**
+     * @Given a :messageType email has NOT been sent to that user
+     */
+    public function anEmailHasNotBeenSentToThatUser(string $messageType)
+    {
+        return;
```
forevermatt: Looks like the body of this function was never written. I think all that it would need to do is delete any emails of that type that were sent to that user.

application/features/bootstrap/EmailContext.php
```diff
@@ -669,6 +736,22 @@ public function iSeeThatAMfaDisabledEmailShouldBeSent()
         assert::true($this->mfaDisabledEmailShouldBeSent);
     }
 
+    /**
+     * @Then I see that a get backup codes email has NOT been sent recently
+     */
+    public function iSeeThatAGetBackupCodesEmailHasNotBeSent()
+    {
+        assert::false($this->getBackupCodesEmailHasBeenSent);
```
forevermatt: Since this would be a simple fix I'll mention it, even though the code apparently works either way:
All of these `assert::` calls should probably be `Assert::` since class names in PHP use TitleCase.

application/features/email.feature
```diff
@@ -213,6 +206,41 @@ Feature: Email
...
       | to send                 | does exist     | does exist      | delete_mfa        | should NOT  |
       | to send                 | does NOT exist | does NOT exist  | delete_mfa        | should      |
-      | Not to send             | does NOT exist | does NOT exist  | delete_mfa        | should NOT  |
+      | NOT to send             | does NOT exist | does NOT exist  | delete_mfa        | should NOT  |
       | to send                 | does NOT exist | does NOT exist  | create_mfa        | should NOT  |
-      | Not to send             | does NOT exist | does NOT exist  | create_mfa        | should NOT  |
\ No newline at end of file
+      | NOT to send             | does NOT exist | does NOT exist  | create_mfa        | should NOT  |
+
+  Scenario Outline: When to send refresh backup codes emails
```
forevermatt: Do you think it would be worth adding a line to this to specify what the fakeEmailer's `$minimumBackupCodesBeforeNag` is set to?

application/features/email.feature
```diff
@@ -213,6 +206,41 @@ Feature: Email
...
+    Examples:
+      | sendRefreshBackupCodesEml | backupCodes    | shouldOrNot |
+      | to send                   | 9              | should NOT  |
+      | to send                   | 5              | should NOT  |
+      | to send                   | 4              | should      |
+      | to send                   | 1              | should      |
+      | NOT to send               | 5              | should NOT  |
+      | NOT to send               | 4              | should NOT  |
+
+  Scenario Outline: Check if an email type has been sent recently (negative days ago means not sent yet)
```
forevermatt: Similarly here, should we add a line setting the `emailRepeatDelayDays` to something specific? That could protect us against the tests failing if we've set our local value for that to something other than the default.

forevermatt:
Other than my latest comments above, I think my only question is whether there is a test which verifies that the new query/iterator code works as expected:
```php
$query = (new Query)->from('user');

// iterate over one user at a time.
foreach ($query->each() as $user) {
    ...
}
```

application/common/components/Emailer.php
Baggerone: Done

application/common/models/Mfa.php
Baggerone: Done

application/features/bootstrap/EmailContext.php
Baggerone: Removed

application/features/bootstrap/EmailContext.php
Baggerone: Done

application/features/bootstrap/EmailContext.php
Baggerone: Done

application/features/email.feature
Baggerone: Done

application/features/email.feature
Baggerone: Done

PR #118, opened 2017-12-21
Updated email content and fixing require_mfa updates

fillup:


fillup:
To be released as `3.3.2`

PR #119, opened 2018-01-05
Move code for emailing get-backup-codes to the validate and remove function

Baggerone:


application/common/components/Emailer.php
```diff
@@ -402,13 +402,13 @@ public function shouldSendGetBackupCodesMessageTo($user)
 
     /**
      * @param User $user
+     * @param int $backupCodeCount - the number of backup codes left for the user (after a deletion)
      * @return bool
      */
-    public function shouldSendRefreshBackupCodesMessageTo($user)
+    public function shouldSendRefreshBackupCodesMessageTo($backupCodeCount)
```
forevermatt: Since we're no longer passing a User to this function, you could remove the "...To" from the end of the function name (since it's not dealing with sending the message "to" anyone in particular). Looks like there's also an `@param` in the doc. comment that could be removed.

application/common/models/MfaBackupcode.php
```diff
@@ -26,12 +26,14 @@ public function rules(): array
      * Check if given value exists, if so delete and return true, else false
      * @param int $mfaId
      * @param int $code
+     * @param bool $sendEmail optional (default=True) Whether to send an email (False is used by tests)
      * @return bool
      * @throws ServerErrorHttpException
      */
-    public static function validateAndRemove(int $mfaId, int $code): bool
+    public static function validateAndRemove(int $mfaId, int $code, $sendEmail=True): bool
```
forevermatt: It looks like we can remove the need for this `$sendEmail` parameter by simply changing the relevant test. If we simply have the `When to send refresh backup codes emails` test remove records of any emails that have been sent as the last step before the `When` (instead of earlier in the list of `Given`s), then the test works as desired and we don't need to add this parameter.

application/features/bootstrap/EmailContext.php
```diff
@@ -566,19 +572,30 @@ public function aBackupCodeMfaOptionDoesNotExist()
...
-        $backupMfa->refresh();
+        $codeCount = count($this->tempBackupCodes);
+
+        Assert::greaterThanEq($codeCount, $desiredCount, 'There are not enough backup codes to run the test.');
+
+        $countDiff = $codeCount - $desiredCount;
+
+        for ($i = 0; $i < $countDiff; $i++) {
+            $nextCode = array_shift($this->tempBackupCodes);
+            MfaBackupcode::validateAndRemove($backupMfa->id, $nextCode, False);
```
forevermatt: You may be able to simplify some of this code by simply blowing away any existing backup codes and re-creating exactly how many you need. That might take less code that looping through and calling `validateAndRemove(...)` for each backupcode that you don't need. Just a suggestion.

forevermatt: Never mind. I played around with this idea a little bit more, and since you need to both (A) have an exact number of backup codes remaining and (B) know what they are, it looks like the way you're currently doing it is better than what I suggested.

application/features/bootstrap/EmailContext.php
```diff
@@ -603,8 +620,9 @@ public function aBackupCodeMfaOptionWasXUsedDaysAgo($lastUsedDaysAgo)
     public function aBackupCodeIsUsedUpByThatUser()
     {
         $backupMfa = $this->getMfa(Mfa::TYPE_BACKUPCODE);
-        $backUpCode = MfaBackupcode::find()->where(['mfa_id' => $backupMfa->id])->one();
-        $backUpCode->delete();
+        $backUpCode = array_shift($this->tempBackupCodes);
+
+        MfaBackupcode::validateAndRemove($backupMfa->id, $backUpCode);
```
forevermatt: One other thing probably worth doing is wrapping this in an `Assert::true(...)` to make sure the backup code was actually removed. I accidentally discovered the need for that when experimenting with these tests.

It may also be worth adding the `Assert::true(...)` where we call `validatedAndRemove(...)` up around line 594.

application/common/components/Emailer.php
Baggerone: Done

application/features/bootstrap/EmailContext.php
Baggerone: I added this Assert but not the previous one, since there was a separate Assert  there to verify the total number of remaining backup codes.

Baggerone: I added the Assert here but not in the earlier place, since it has a following Assert for the number of remaining codes.

application/common/models/MfaBackupcode.php
Baggerone: Done

PR #122, opened 2018-01-08
Feature/stop casting codes to int

Baggerone:


forevermatt:
@fillup It just occurred to me that this change will cause any already-generated backup codes that start with a zero (about 10% of all backup codes) to no longer work. Should we simply leave the code as it was before (so that backup codes are cast to `int` before being hashed and saved to the database, as well as before being checked against the hashes in the database)?

Before now, a backup code of `'12345678'` would have been converted to an int (`12345678`), hashed, and then the hash saved to the database. When someone tried to use that backup code, they would type in `'12345678'`, which would be converted to an int (`12345678`) and that would be (implicitly cast back to a string and) would match the hash.

The problem is that for codes that start with a `0` (such as `'01234567'`), it would be cast to an int (`1234567`), then hashed and saved. It still worked up until now because, when someone typed it in (which we receive as a string `'01234567'`), it would be cast to an int (`1234567`), which is then implicitly cast to a string and checked against the hash (and it would match).

This fix was to stop casting to int, so that new codes would keep any leading zeros, and codes people type in from now on would also keep their leading zeros... which unfortunately means those codes would _not_ match the hashes created before now (back when zeros were being stripped off before the hash was created).

Should we simply leave the code as it was, allowing leading zeros to be stripped, since that basically is just using implied leading zeros for any backup codes shorter than 8 characters? While not stripping the leading zeros seems more "correct", I don't think stripping off the leading zeros is actually decreasing security in any way, and "fixing" that now would break a tenth of the backup codes our users already have.

Thoughts?

PR #124, opened 2018-01-09
Allow for missing leading zeros in backup codes

Baggerone:


application/common/models/MfaBackupcode.php
```diff
@@ -31,10 +31,26 @@ public function rules(): array
      */
     public static function validateAndRemove(int $mfaId, $code): bool
     {
+        $intCode = intval($code);
+        $codeStartsWithZero = (! empty($code) && $code[0] === '0');
```
forevermatt: The code...
`(! empty($code) && $code[0] === '0')`
... is complicated enough, I have to look closely and think for a bit to decide whether the `!` applies to the `empty()` or to the result of the `&&`:
`((! empty($code)) && $code[0] === '0')`
vs.
`(! (empty($code) && $code[0] === '0'))`

What would you think of simplifying it to something like this?
`substr($code, 0, 1) === '0'`
That works even if an empty string was given, though it's possible I overlooked some different shortcoming.

Thoughts?

application/features/bootstrap/MfaUnitTestsContext.php
```diff
@@ -0,0 +1,137 @@
...
+
+        MfaBackupcode::insertBackupCode($this->mfaId, $fullBackupCode);
+    }
+
+    /**
+     * @When a matching backup code is provided for validation
+     */
+    public function aMatchingBackupCodeIsProvidedForValidation()
+    {
+        $this->backupCodeWasValid = MfaBackupcode::validateAndRemove($this->mfaId, $this->inputBackupCode);
```
forevermatt: Very minor issue, feel free to ignore it, but this line is long enough that for readability (especially in diffs and other situations where lines much longer than 80 chars get wrapped), it may be worth breaking these parameters out onto their own lines:
```php
$this->backupCodeWasValid = MfaBackupcode::validateAndRemove(
    $this->mfaId,
    $this->inputBackupCode
);
```
Just a suggestion.

application/common/models/MfaBackupcode.php
Baggerone: Fixed.

PR #126, opened 2018-01-11
Feature/fix bug in email

Baggerone:


application/common/models/Mfa.php
```diff
@@ -66,12 +66,40 @@ public function fields(): array
...
-        if ($insert) {
-            $this->sendAppropriateMessages($this->user, self::EVENT_TYPE_CREATE);
+        /*
+         *   Send an "Mfa Added email" for a new backup code option and for
+         * other types of mfa options that are newly verified
+         *
+         *   Don't send emails before they are verified, since the email will
+         * not include the most recently added option.
+         */
+        if ($this->isNewlyVerified($insert, $changedAttributes)) {
```
forevermatt: I find myself jumping back and forth between this code (in `afterSave()`) and the contents of the `sendAppropriateMessages()` function to check the usage of this `isNewlyVerified()` function call.

It feels like we're letting the logic of when to send the various emails leak outside of their respective `$emailer->shouldSend....()` methods. What would you think of having `afterSave()` always call `$this->sendAppropriateMessages(...)`, and we could keep all of the conditions (for when to send each email) cleanly contained within those `shouldSend...()` methods?

I know that means `afterSave()` will need to pass along more parameters to `sendAppropriateMessages()`, but that seems an acceptable cost.

application/features/bootstrap/MfaUnitTestsContext.php
```diff
@@ -49,19 +57,84 @@ protected function createNewUserInDatabase($username)
...
+        }
+        $mfa = new Mfa();
+        $mfa->user_id = $user->id;
+        $mfa->type = $type;
+        $mfa->verified = $verified;
+
+        Assert::true($mfa->save(), "Could not create new mfa.");
+        $user->refresh();
+        $this->mfaId = $mfa['id'];
+        $this->mfa = Mfa::findOne(['id' => $mfa['id']]);
```
forevermatt: It looks like this line...
```php
$this->mfa = Mfa::findOne(['id' => $mfa['id']]);
```
... is simply re-getting the Mfa that we just created. Would the following have worked?
```php
$this->mfa = $mfa;
```

application/features/bootstrap/MfaUnitTestsContext.php
```diff
@@ -49,19 +57,84 @@ protected function createNewUserInDatabase($username)
...
+        $this->createMfa(Mfa::TYPE_BACKUPCODE);
         MfaBackupcode::deleteAll();
     }
 
+    /**
+     * @Given I have a user with an unverified totp mfa option
+     */
+    public function iHaveAUserWithAnUnverifiedTotpMfaOption()
+    {
+        Mfa::deleteAll([]);
```
forevermatt: In case you find this preferable, when using `deleteAll()` to delete all of those records (no conditions), you don't need to pass it any parameters at all.

application/common/models/Mfa.php
```diff
@@ -66,12 +66,40 @@ public function fields(): array
...
+    /**
+     * Whether this is both a new Mfa instance and already verified
+     *   (basically just for a new backup code Mfa option)
+     *  -- OR --
+     * Whether the Mfa option both had its verified value changed and is now verified
+     *
+     * @param array $changedAttributes
+     * @return bool
+     */
+    public function isNewlyVerified($insert, $changedAttributes)
```
forevermatt: Missing `@param` for `$insert` parameter.

application/features/bootstrap/MfaUnitTestsContext.php
Baggerone: Done.

application/features/bootstrap/MfaUnitTestsContext.php
Baggerone: Done.

application/common/models/Mfa.php
forevermatt: Never mind. That you helped me see the full significance of the `EVENT_TYPE_VERIFY`, I think this is fine.

PR #130, opened 2018-01-18
Include MFA rate limit email content

fillup:


fillup:
Will be released as `3.4.1`

PR #131, opened 2018-02-28
Feature/dont email unverified

Baggerone:
Don't send emails when an unverified Mfa option is deleted

### DIFF
--- a/application/common/components/Emailer.php
+++ b/application/common/components/Emailer.php
@@ -493,30 +493,32 @@ class Emailer extends Component
     }
 
     /**
-     * Whether the user just deleted an MFA option and there is still one or more left
+     * Whether the user just deleted a verified MFA option and there is still one or more left
      *
      * @param User $user (assumes the user instance has already been refreshed)
      * @param string Mfa::EVENT_TYPE_*
      * @return bool
      */
-    public function shouldSendMfaOptionRemovedMessageTo($user, $mfaEventType)
+    public function shouldSendMfaOptionRemovedMessageTo($user, $mfaEventType, $mfa)
     {
         return $this->sendMfaOptionRemovedEmails
             && $mfaEventType === Mfa::EVENT_TYPE_DELETE
+            && $mfa->verified
             && count($user->getVerifiedMfaOptions()) > 0;
     }
 
     /**
-     * Whether the user has just deleted the last MFA option
+     * Whether the user has just deleted the last verified MFA option
      *
      * @param User $user (assumes the user instance has already been refreshed)
      * @param string Mfa::EVENT_TYPE_*
      * @return bool
      */
-    public function shouldSendMfaDisabledMessageTo($user, $mfaEventType)
+    public function shouldSendMfaDisabledMessageTo($user, $mfaEventType, $mfa)
     {
         return $this->sendMfaDisabledEmails
             && $mfaEventType === Mfa::EVENT_TYPE_DELETE
+            && $mfa->verified
             && count($user->getVerifiedMfaOptions()) < 1;
     }
     
--- a/application/common/components/Emailer.php
+++ b/application/common/components/Emailer.php
@@ -497,6 +497,7 @@ class Emailer extends Component
      *
      * @param User $user (assumes the user instance has already been refreshed)
      * @param string Mfa::EVENT_TYPE_*
+     * @param Mfa $mfa
      * @return bool
      */
     public function shouldSendMfaOptionRemovedMessageTo($user, $mfaEventType, $mfa)
@@ -512,6 +513,7 @@ class Emailer extends Component
      *
      * @param User $user (assumes the user instance has already been refreshed)
      * @param string Mfa::EVENT_TYPE_*
+     * @param Mfa $mfa
      * @return bool
      */
     public function shouldSendMfaDisabledMessageTo($user, $mfaEventType, $mfa)
--- a/application/features/email.feature
+++ b/application/features/email.feature
@@ -71,7 +71,7 @@ Feature: Email
       And I remove records of any emails that have been sent
       And no mfas exist
       And a user already exists
-      And a u2f mfa option <u2fExistsOrNot>
+      And a verified u2f mfa option <u2fExistsOrNot>
       And a backup code mfa option was used <backupUsedDaysAgo> days ago
       And a totp mfa option was used <totpUsedDaysAgo> days ago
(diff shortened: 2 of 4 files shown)


application/common/components/Emailer.php
```diff
@@ -493,30 +493,32 @@ public function shouldSendMfaEnabledMessageTo($user, $mfaEventType)
     }
 
     /**
-     * Whether the user just deleted an MFA option and there is still one or more left
+     * Whether the user just deleted a verified MFA option and there is still one or more left
      *
      * @param User $user (assumes the user instance has already been refreshed)
      * @param string Mfa::EVENT_TYPE_*
      * @return bool
```
forevermatt: Looks like you'll also want to add an `@param` on each of these functions for the new `$mfa` parameter.

application/features/email.feature
```diff
@@ -185,7 +185,6 @@ Feature: Email
       | to send                 | does exist     | does NOT exist  | delete_mfa        | should      |
       | to send                 | does NOT exist | does exist      | delete_mfa        | should      |
       | to send                 | does exist     | does exist      | delete_mfa        | should      |
-      | to send                 | does NOT exist | does NOT exist  | delete_mfa        | should NOT  |
```
forevermatt: Out of curiosity, why was this example removed from the test?

application/features/email.feature
```diff
@@ -205,11 +204,38 @@ Feature: Email
       | to send            | does exist     | does NOT exist  | delete_mfa   | should NOT  |
       | to send            | does NOT exist | does exist      | delete_mfa   | should NOT  |
       | to send            | does exist     | does exist      | delete_mfa   | should NOT  |
-      | to send            | does NOT exist | does NOT exist  | delete_mfa   | should      |
```
forevermatt: Same here (I'm curious why this test case was removed).

application/features/email.feature
```diff
@@ -205,11 +204,38 @@ Feature: Email
...
       | to send            | does NOT exist | does NOT exist  | verify_mfa   | should NOT  |
       | NOT to send        | does NOT exist | does NOT exist  | verify_mfa   | should NOT  |
 
+
+  Scenario Outline: When to send an email after a verified u2f mfa option has been deleted.
+    Given we are configured to send mfa disabled emails
+      And we are configured to send mfa option removed emails
+      And no mfas exist
+      And a user already exists
+      And a verified u2f mfa option used to exist
```
forevermatt: What does it mean that a verified U2F MFA option "used to exist"?

application/features/email.feature
```diff
@@ -205,11 +204,38 @@ Feature: Email
...
+      And the latest mfa event type was delete_mfa
+    When I check if a mfa <optionRemovedOrDisabled> email should be sent
+    Then I see that a mfa <optionRemovedOrDisabled> email <shouldOrNot> be sent
+
+    Examples:
+      | optionRemovedOrDisabled | shouldOrNot |
+      | option removed          | should NOT  |
```
forevermatt: I think there's a typo here: when we have just deleted a verified MFA option, then we _should_ send the "MFA option removed" email. Right?

application/features/email.feature
```diff
@@ -205,11 +204,38 @@ Feature: Email
...
+      | disabled                | should      |
+
+  Scenario: When to send an email after an unverified u2f mfa option has been deleted.
+    Given we are configured to send mfa disabled emails
```
forevermatt: Whether we are configured to send MFA-disabled emails seems irrelevant to this test (since the When/Then only care about MFA-option-removed emails), so this line could be removed and the `Given` moved down to the next line.

application/features/email.feature
```diff
@@ -205,11 +204,38 @@ Feature: Email
...
+      And we are configured to send mfa option removed emails
+      And no mfas exist
+      And a user already exists
+      And an unverified u2f mfa option does exist
+      And a verified u2f mfa option does exist but not as the test mfa
+      And the latest mfa event type was delete_mfa
+    When I check if a mfa option removed email should be sent
```
forevermatt: Based on the wording of this test scenario, I don't think it's describing the situation we're trying to test here. (For example, it doesn't seem to say whether the MFA option just deleted was verified or not... simply that there is an unverified MFA option and a verified MFA option that do exist).

What if we changed from these lines...
```gherkin
    And an unverified u2f mfa option does exist
    And a verified u2f mfa option does exist but not as the test mfa
    And the latest mfa event type was delete_mfa
```
... to this:
```gherkin
    And a verified u2f mfa option does exist
    And an unverified u2f mfa option was just deleted
```

application/features/email.feature
Baggerone: It was moved down to a separate test, because this test doesn't have a way of saying that  the Mfa used to exist but doesn't anymore.  It just says that right now, it doesn't exist.  So, it didn't have access to a Mfa instance to send to the Emailer method.

application/features/email.feature
Baggerone: It gives the test a way to hold on to a Mfa instance to send to the Emailer class. It doesn't save it to the database and then delete it, since deleting it isn't doable with the external service we use.

application/features/email.feature
Baggerone: When it's the last mfa option of that type, we only send the mfa disabled email and not the mfa option removed email.

forevermatt: I don't think we differentiate by MFA type when deciding whether to sent the "MFA option removed" vs. the "MFA disabled" emails: if they have _any_ remaining MFA options, then they have simply removed an MFA option; if they don't have any MFA options remaining, then they have disabled MFA.

forevermatt: Never mind, you're right. I was overlooking that this was testing which email to send when the _last_ verified MFA option is deleted.

application/features/email.feature
```diff
@@ -205,11 +204,35 @@ Feature: Email
...
+      And a verified u2f mfa option was just deleted
+    When I check if a mfa <optionRemovedOrDisabled> email should be sent
+    Then I see that a mfa <optionRemovedOrDisabled> email <shouldOrNot> be sent
+
+    Examples:
+      | optionRemovedOrDisabled | shouldOrNot |
+      | option removed          | should NOT  |
+      | disabled                | should      |
+
+  Scenario: When to send an email after an unverified u2f mfa option has been deleted.
```
forevermatt: This test scenario looks fine now. Should we also have one like this but for testing the sending of the MFA-disabled email?

PR #139, opened 2018-10-24
Add recovery method nag

Schparky:
This is the first of breaking changes merged into develop-4.0 branch.
* New migration to add nag_for_method_after date on user table
* New method object on user resource, returned from /authentication and /user endpoints
* POST /authentication checks review dates and updates by priority mfa then method

NOTE: Users with existing methods will get a nag. Need to bring methods into broker to fix this.

### DIFF
--- a/application/common/config/main.php
+++ b/application/common/config/main.php
@@ -193,7 +193,7 @@ return [
         'idpName'                       => $idpName,
         'idpDisplayName'                => $idpDisplayName,
         'mfaNagInterval'                => Env::get('MFA_NAG_INTERVAL', '+30 days'),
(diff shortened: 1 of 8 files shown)


application/common/config/main.php
```diff
@@ -193,6 +193,7 @@
         'idpName'                       => $idpName,
         'idpDisplayName'                => $idpDisplayName,
         'mfaNagInterval'                => Env::get('MFA_NAG_INTERVAL', '+30 days'),
+        'methodNagInterval'             => Env::get('METHOD_NAG_INTERVAL', '+30 days'),
```
fillup: Can we change the default to `+6 months`?
```suggestion
        'methodNagInterval'             => Env::get('METHOD_NAG_INTERVAL', '+6 months'),
```

PR #140, opened 2018-10-31
Moved recovery methods from idp-pw-api

Schparky:
Moved and refactored the Method and MethodController classes from idp-pw-api. All functionality is present except for the verification email, which will be implemented in another PR.

### DIFF
--- a/api.raml
+++ b/api.raml
@@ -153,14 +153,6 @@ schemas:
           "spouse_email": {
             "required": false,
             "type": "string"
-          },
-          "mfa_review_after": {
-            "required": false,
-            "type": "string"
-          },
-          "method_review_after": {
-            "required": false,
-            "type": "string"
           }
         },
         "type": "object"
--- a/api.raml
+++ b/api.raml
@@ -360,6 +360,39 @@ schemas:
         },
         "type": "object"
       }
+  - MethodResponse: |
+    {
+        "$schema": "http://json-schema.org/draft-03/schema",
+        "properties": {
+          "uid": {
+            "required": true,
+            "type": "string"
+          },
+          "value": {
+            "required": true,
+            "type": "string"
+          }
+        },
+        "type": "object"
+    }
+  - MethodListResponse: |
+    {
+        "$schema": "http://json-schema.org/draft-03/schema",
+        "items": {
+          "properties": {
+            "uid": {
+              "required": true,
+              "type": "string"
+            },
+            "value": {
+              "required": true,
+              "type": "string"
+            }
+          },
+          "type": "object"
+        }
+        "type": "array"
+    }
 
 /authentication:
   post:
@@ -721,6 +754,7 @@ schemas:
           200:
             body:
               application/json:
+                schema: MethodListResponse
                 example: |
                   [
                       {
@@ -942,3 +976,27 @@ schemas:
           429:
             description: |
               Too many failed attempts for this MFA record recently.
+/method:
+  /{uid}:
+    get:
+      description: Retrieve a specific recovery method.
+      body:
+        application/json:
+          example: |
+            {
+              "employee_id": "12345"
+            }
+      responses:
+        200:
+          description: Recovery method found
+            body:
+              application/json:
+                schema: MethodResponse
+                example: |
+                  {
+                    "uid": "12345678901234567890123456789012",
+                    "value": "email@example.org"
+                  }
+        404:
+        500:
+
--- a/api.raml
+++ b/api.raml
@@ -714,6 +714,20 @@ schemas:
                           "data": []
                       }
                   ]
+    /method:
+      get:
+        description: Get a list of verified recovery methods for user
+        responses:
+          200:
+            body:
+              application/json:
+                example: |
+                  [
+                      {
+                          "uid": "0YrSiNp2P6sfkleepNNvFYMQWuLfssve",
+                          "value": email@example.org
+                      },
+                  ]
     /password:
       put:
         description: Set the password for a User.
--- a/api.raml
+++ b/api.raml
@@ -998,6 +998,8 @@ schemas:
                 "uid": "12345678901234567890123456789012",
                 "value": "email@example.org"
               }
+      400:
+        description: Missing parameter
       409:
         description: New method matches an already existing verified method
       500:
--- a/api.raml
+++ b/api.raml
@@ -751,6 +751,7 @@ schemas:
       get:
         description: Get a list of verified recovery methods for user
         responses:
+          400:
           200:
             body:
               application/json:
@@ -998,5 +999,28 @@ schemas:
                     "value": "email@example.org"
                   }
         404:
+          description: User not found or method not found
+    delete:
+      description: Delete a specific recovery method.
+      body:
+        application/json:
+          example: |
+            {
+              "employee_id": "12345"
+            }
+      responses:
+        200:
+          description: Recovery method found
+            body:
+              application/json:
+                schema: MethodResponse
+                example: |
+                  {
+                    "uid": "12345678901234567890123456789012",
+                    "value": "email@example.org"
+                  }
+        404:
+          description: User not found or method not found
         500:
+          description: Delete failed
 
--- a/api.raml
+++ b/api.raml
@@ -1008,7 +1008,7 @@ schemas:
         description: Database or other system error
   /{uid}:
     get:
-      description: Retrieve a specific, validated recovery method.
+      description: Retrieve a specific, verified recovery method.
       body:
         application/json:
           example: |
@@ -1074,7 +1074,7 @@ schemas:
             description: Database or other system error
     /resend:
       put:
-        description: Resend the recovery email
+        description: Resend the method verification email
         body:
           application/json:
             example: |
--- a/api.raml
+++ b/api.raml
@@ -979,13 +979,15 @@ schemas:
               Too many failed attempts for this MFA record recently.
 /method:
   post:
-    description: Create a new recovery method and send verification email.
+    description: >
+      Create a new recovery method and send verification email. At this point,
+      all methods are email, so 'value' is always an email address.
     body:
       application/json:
         example: |
           {
             "employee_id": "12345",
-            "value": "987654"
+            "value": "email@example.org"
           }
     responses:
       200:
@@ -1035,16 +1037,8 @@ schemas:
               "employee_id": "12345"
             }
       responses:
-        200:
-          description: Recovery method found
-          body:
-            application/json:
-              schema: MethodResponse
-              example: |
-                {
-                  "uid": "12345678901234567890123456789012",
-                  "value": "email@example.org"
-                }
+        204:
+          description: Recovery method found and deleted
         404:
           description: User not found or method not found
         500:
@@ -1074,6 +1068,8 @@ schemas:
             description: Missing or invalid code
           404:
             description: User not found or method not found
+          429:
+            description: Too many verification attempts
           500:
             description: Database or other system error
     /resend:
@@ -1092,4 +1088,3 @@ schemas:
             description: Method already verified
           404:
             description: User not found or method not found
-
--- a/api.raml
+++ b/api.raml
@@ -978,6 +978,30 @@ schemas:
             description: |
               Too many failed attempts for this MFA record recently.
 /method:
+  post:
+    description: Create a new recovery method and send verification email.
+    body:
+      application/json:
+        example: |
+          {
+            "employee_id": "12345",
+            "value": "987654"
+          }
+    responses:
+      200:
+        description: Recovery method created and email sent.
+        body:
+          application/json:
+            schema: MethodResponse
+            example: |
+              {
+                "uid": "12345678901234567890123456789012",
+                "value": "email@example.org"
+              }
+      409:
+        description: New method matches an already existing verified method
+      500:
+        description: Database or other system error
   /{uid}:
     get:
       description: Retrieve a specific recovery method.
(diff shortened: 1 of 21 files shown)


api.raml
```diff
@@ -928,3 +977,119 @@ schemas:
...
               Too many failed attempts for this MFA record recently.
+/method:
+  post:
+    description: Create a new recovery method and send verification email.
+    body:
+      application/json:
+        example: |
+          {
+            "employee_id": "12345",
+            "value": "987654"
```
forevermatt: This example of what to `POST` to create a new method seems confusing to me. How would an `employee_id` and a numeric-string `value` be enough for the server to know what to create for the new method?

api.raml
```diff
@@ -928,3 +977,119 @@ schemas:
...
+    responses:
+      200:
+        description: Recovery method created and email sent.
+        body:
+          application/json:
+            schema: MethodResponse
+            example: |
+              {
+                "uid": "12345678901234567890123456789012",
+                "value": "email@example.org"
```
forevermatt: What is the meaning/significance of the `value` in the MethodResponse?

api.raml
```diff
@@ -928,3 +977,119 @@ schemas:
...
+                  }
+          400:
+            description: Missing or invalid code
+          404:
+            description: User not found or method not found
+          500:
+            description: Database or other system error
+    /resend:
+      put:
+        description: Resend the recovery email
```
forevermatt: This could also be an SMS (not just an email), right? If so, that might be worth including in the docs here:
`description: Resend the recovery email/text` or `description: Resend the recovery message`

application/common/config/main.php
```diff
@@ -202,6 +202,7 @@
         'googleAnalytics'               => [
             'trackingId' => Env::get('GA_TRACKING_ID'),
             'clientId'   => Env::get('GA_CLIENT_ID'),
-        ]
+        ],
+        'reset' => ENV::getArrayFromPrefix('RESET_'),
```
forevermatt: The class name here should probably be changed from `ENV` to `Env`.

application/common/helpers/Utils.php
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * Return a random string of numbers
+     * @param int $length [default=4]
+     * @return string
+     * @throws \Exception
+     */
+    public static function getRandomDigits($length = 4)
+    {
+        $result = '';
+        while (strlen($result) < $length) {
```
forevermatt: Now that we're using PHP >= 7, this whole thing could probably be simplified to instead use [random_int()](http://php.net/random_int). Something like this might work:
```php
$result = '';
for ($i = 0; $i < $length; $i++) {
    $result .= random_int(0, 9);
}
return $result;
```

api.raml
Schparky: It's the same as the pw-api POST/method except that type is not required since 'email' is the only type.

api.raml
Schparky: `value` is always an email, as of today. I'll make this more clear.

api.raml
Schparky: As of today, recovery methods are always email addresses. I'll make that note in the RAML file.

application/common/models/Method.php
```diff
@@ -0,0 +1,179 @@
...
+        if ($this->isVerified()) {
+            throw new ConflictHttpException('Method already verified', 1540689804);
+        }
+
+        $this->verification_attempts++;
+        if ( ! $this->save()) {
+            throw new ServerErrorHttpException('Save error after incrementing attempts', 1461441850);
+        }
+
+//        $this->sendVerificationEmail();
```
forevermatt: Should this be commented out?

application/common/models/Method.php
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    /**
+     * Send a verification email message
+     */
+    public function sendVerificationEmail()
+    {
+        Verification::sendEmail(
```
forevermatt: The `Verification` class doesn't seem to exist.

application/common/helpers/Utils.php
Schparky: Thanks!

application/common/models/Method.php
Schparky: For now, yes. It's not implemented yet.

application/common/models/Method.php
Schparky: Yup. TBD.

application/console/migrations/m181025_155801_create_method_table.php
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Handles the creation of table `method`.
+ */
+class m181025_155801_create_method_table extends Migration
+{
+    /**
+     * @inheritdoc
+     */
+    public function up()
```
forevermatt: I don't know if it will actually make a difference in this case, but I think our general practice is to use `safeUp()`/`safeDown()` rather than `up()`/`down()` in case there is some benefit to doing the changes as one transaction.

application/console/migrations/m181025_155801_create_method_table.php
```diff
@@ -0,0 +1,43 @@
...
+                'value' => 'varchar(255) not null',
+                'verified' => 'boolean not null default false',
+                'verification_code' => 'varchar(64) null',
+                'verification_attempts' => 'smallint null',
+                'verification_expires' => 'datetime null',
+                'created' => 'datetime not null',
+            ],
+            "ENGINE=InnoDB DEFAULT CHARSET=utf8"
+        );
+        $this->addForeignKey('fk_method_user_id','{{method}}','user_id','{{user}}','id','CASCADE','NO ACTION');
```
forevermatt: @fillup, what did we decide to do regarding CASCADE vs. NO ACTION for on-delete and on-update? I think we've been using the Yii model's `beforeDelete()` to individually delete child records (since that lets us enforce any `beforeDelete()` business rules for the child records), but I'm not certain.

application/console/migrations/m181025_155801_create_method_table.php
```diff
@@ -0,0 +1,43 @@
...
+        $this->createIndex('uq_method_user_type_value','{{method}}',['user_id','value'],true);
+        $this->createIndex('uq_method_verification_code','{{method}}','verification_code',true);
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function down()
+    {
+        $this->dropTable('{{method}}');
```
forevermatt: Have you tested applying this down-migration? I'm wondering whether the `dropTable()` will also automatically drop the indexes and foreign keys, or whether they need to be individually dropped here before the call to `dropTable()`.

application/features/bootstrap/MethodContext.php
```diff
@@ -0,0 +1,58 @@
...
+use common\models\Method;
+use common\models\User;
+use Webmozart\Assert\Assert;
+
+class MethodContext extends \FeatureContext
+{
+    protected $methodFromDb;
+
+    /**
+     * @Given /^user with employee id (.*) has (?:a|an) (verified|unverified) Method$/
```
forevermatt: This is just a suggestion, but these `@Given` (etc.) DocBlock tags have two different syntaxes: regex-based (which PhpStorm generates by default) and a simpler string matching syntax (which Behat generates by default). I find the simpler syntax preferable in most situations.

You're free to use whichever syntax you find works best in any given situation, but I wanted you to at least know about the simpler syntax. For example, the simpler version for this method would be something like this:
```
@Given user with employee id :employeeId has a(n) :verified Method
```

application/features/method.feature
```diff
@@ -0,0 +1,85 @@
...
+    Then the response status code should be 200
+    And the following data is returned:
+      | property      | value                 |
+      | value         | user123@example.org   |
+    And a method record exists with a value of "user123@example.org"
+    And the following method data should be stored:
+      | property              | value                 |
+      | value                 | user123@example.org   |
+      | verified              | 0                     |
+      | verification_attempts | 1                     |
```
forevermatt: I may have simply misread this test, but when we have simply created a method, why should `verification_attempts` already be at `1`?

application/features/method.feature
```diff
@@ -0,0 +1,85 @@
...
+   Then the response status code should be 200
+   And a method record exists with a value of "unverified@example.com"
+   And the following method data should be stored:
+     | property              | value                 |
+     | verified              | 0                     |
+#   And an email is sent to "unverified@example.com"
+
+ Scenario: Verify a Method
+   Given user with employee id 123 has an unverified Method
+   When I send a "PUT" to "/method/{uid}/verify" with a valid uid
```
forevermatt: It seems unclear in this test how the method was verified. Isn't a `verify` call required to provide the correct `verification_code`?

application/common/models/Method.php
```diff
@@ -0,0 +1,179 @@
...
+     * @throws ConflictHttpException
+     * @throws ServerErrorHttpException
+     */
+    public function sendVerification()
+    {
+        if ($this->isVerified()) {
+            throw new ConflictHttpException('Method already verified', 1540689804);
+        }
+
+        $this->verification_attempts++;
```
forevermatt: Sorry to be rehashing existing design, by why does sending a verification email update the `verification_attempts` counter? Isn't that supposed to be the number of times someone has submitted a value in an attempt to verify the Method?

application/frontend/controllers/MethodController.php
```diff
@@ -0,0 +1,209 @@
...
+
+        if ($method === null) {
+            $method = new Method;
+            $method->user_id = $userId;
+            $method->value = mb_strtolower($value);
+        }
+
+        $method->sendVerification();
+
+        if ( ! $method->save()) {
```
forevermatt: We may want to make this `$method->save()` call _before_ we call `$method->sendVerification()`, since calling that `sendVerification()` method itself does a save call (which is not obvious when reading this code).

It also just seems like we would create and save the method before sending the user an email for them to verify the method (which, if the save failed, won't exist).

application/console/migrations/m181025_155801_create_method_table.php
Schparky: I have now. Seems to work as it is.

application/features/bootstrap/MethodContext.php
Schparky: Thanks. I think I tried the simpler one but must have done something wrong so I thought I needed the regex version.

application/features/method.feature
Schparky: Good question. When I first created the test, I set the check at zero. It failed because sending the initial verification email increments the value. This is the original functionality. Not sure why, though.

application/features/method.feature
Schparky: Not sure I understand your question, but it's probably a poorly written `feature`. The verify call in line 69 does the verify. The `code` is retained in the context definition of the previous line and then sent by the verify call. Do you have any ideas for a more clear feature description?

application/common/models/Method.php
Schparky: I think I agree. Only thing I can think of is the meaning given to `maxAttempts` is defined this way. Who sets the instance-specific configuration, and what information do they use to understand the meaning of the variables? I will go ahead and remove this line and change the test accordingly, but if you or @fillup have any additional background information, I'll change it back.

Schparky:
@forevermatt, I have made a few code edits and replied to your review comments. I 'resolved' those that were simple and left a few open for discussion.

application/features/method.feature
forevermatt: I think this would be clearer:
```suggestion
   When I send the correct code to verify that Method
```

Part of the problem, I think, is that the `Given user with employee id 123 has an unverified Method` step is doing more than it claims to: it's also inserting the correct `verification_code` into the request body for whatever request is made next.

Instead, I'd recommend we wait to add that code to the request until we're in a step that specifically mentions it. As an example, we could remove the ` $this->iChangeThe('code', $method->verification_code)` line from `MethodContext->userHasAMethod(...)` and instead use something like the following for `When I send the correct code to verify that Method`:
```php
    /**
     * @When I send the correct code to verify that Method
     */
    public function iSendTheCorrectCodeToVerifyThatMethod()
    {
        $method = Method::findOne(['uid' => $this->tempUid]);
        $this->iChangeThe('code', $method->verification_code);
        $this->iSendAToWithAValidUid('PUT', '/method/{uid}/verify');
    }
```

application/common/models/Method.php
fillup: We may have it this way so if someone kept switching between methods without attempting to verify it we'd at least stop it from becoming a spam bot. We can have a reasonably high limit to ensure the ability to switch methods and click links, but why not prevent them from switching 200 times and generating a lot of emails?

application/console/migrations/m181025_155801_create_method_table.php
Schparky: I added cascading delete code to `beforeDelete()` in the User model, and changed the `$delete` parameter of `addForeignKey` to `NO ACTION`.

application/features/method.feature
Schparky: I implemented your suggestions. Thank you.

application/common/models/Method.php
Schparky: I will leave the code as it was, then, so that the attempts counter will increment when the email is sent as well as when verification attempts are made.

application/common/models/User.php
```diff
@@ -54,7 +54,20 @@ public function beforeDelete()
                 return false;
             }
         }
-        
+
+        foreach ($this->methods as $method) {
+            if (! $mfa->delete()) {
```
forevermatt: Looks like a typo here
```suggestion
            if (! $method->delete()) {
```

PR #142, opened 2018-11-16
Change HTTP status code for password reuse error

Schparky:
Need to distinguish a password reuse error from other validation errors. Changing the HTTP status code from 422 to 409 in this situation.

### DIFF
--- a/api.raml
+++ b/api.raml
@@ -828,7 +828,7 @@ schemas:
                     "name": "Conflict",
                     "message": "May not be reused yet.",
                     "code": 1542395933,
-                    "status": 422
+                    "status": 409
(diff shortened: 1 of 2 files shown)


api.raml
```diff
@@ -819,6 +819,17 @@ schemas:
...
+          409:
+            description: A recently-used password was given.
+              application/json:
+                schema: Error
+                example: |
+                  {
+                    "name": "Conflict",
+                    "message": "May not be reused yet.",
+                    "code": 1542395933,
+                    "status": 422
```
wcjr:
```suggestion
                    "status": 409
```

PR #143, opened 2018-11-16
Include unverified methods in list

Schparky:
In the `/user/{employee_id}/method` endpoint, we want unverified methods as well as verified ones. Removing the `verified` filter and adding a `verified` property to the response json object.

### DIFF
--- a/api.raml
+++ b/api.raml
@@ -1,9 +1,10 @@
-#%RAML 0.8
+#%RAML 1.0
 title: ID Broker API
+version: 4
 protocols: [ HTTPS ]
 mediaType: application/json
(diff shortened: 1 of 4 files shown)


Schparky:
@wcjr do you want the individual `GET /method/{id}` call to filter out unverified recovery methods? At present it does, so if you ask for an unverified method, you will get a 404.

wcjr:
do we have a mechanism in the cron that cleans up unverified methods after some duration?

Schparky:
Not a cron task, but every call to `createMethod` or `verifyMethod` cleans out expired, unverified methods.

PR #144, opened 2018-11-26
Implement method verification emails.

Schparky:
Recovery method verification emails are now being sent by the email service configured in broker. This is the last bit of method functionality to be moved from pw-api.


application/common/components/Emailer.php
```diff
@@ -30,7 +30,7 @@ class Emailer extends Component
     const SUBJECT_MFA_OPTION_REMOVED_DEFAULT = 'A 2-Step Verification option was removed from your %s account';
     const SUBJECT_MFA_ENABLED_DEFAULT = '2-Step Verification was enabled on your %s account';
     const SUBJECT_MFA_DISABLED_DEFAULT = '2-Step Verification was disabled on your %s account';
-
+    const SUBJECT_METHOD_VERIFY_DEFAULT = 'Verification required - New account recovery method added';
```
forevermatt: If possible, it may be worth rewording this subject to be less computer-ese and more normal-person-English. Perhaps something like "Please verify your new recovery method", but feel free to further tweak that.

Schparky: Good call. I'll do that.

application/common/components/Emailer.php
```diff
@@ -190,16 +192,25 @@ public function getSiteStatus()
...
     }
-    
-    protected function getViewForMessage(string $messageType, string $format)
+
+    /**
+     * @param string $messageType
+     * @param string $format
+     * @return string
+     */
+    protected function getViewForMessage(string $messageType, string $format): string
```
forevermatt: I'm fine with having this PHPDoc block on this method, but what are your thoughts about what to do when the PHPDoc is only a duplication of the data in the method signature? I think I lean towards either removing it completely, or adding a description (of the purpose of the function) to the PHPDoc block. You're welcome to change it, remove it, or leave it as-is.

Schparky: I remember smelling that, but unfortunately did nothing about it. I'm definitely opposed to duplication of information, but something about leaving a DocBlock off entirely doesn't sit well with me. I'll add descriptive text so it adds value.

application/common/config/main.php
```diff
@@ -94,6 +94,7 @@
             'subjectForMfaOptionRemoved' => Env::get('SUBJECT_FOR_MFA_OPTION_REMOVED'),
             'subjectForMfaEnabled' => Env::get('SUBJECT_FOR_MFA_ENABLED'),
             'subjectForMfaDisabled' => Env::get('SUBJECT_FOR_MFA_DISABLED'),
+            'subjectForMethodVerify' => Env::get('SUBJECT_FOR_METHOD_VERIFICATION'),
```
forevermatt: For naming consistency, what would you think about changing `SUBJECT_FOR_METHOD_VERIFICATION` to `SUBJECT_FOR_METHOD_VERIFIFY`?

application/common/models/Method.php
```diff
@@ -177,4 +174,33 @@ public static function deleteExpiredUnverifiedMethods()
...
+     * @return Method
+     * @throws ConflictHttpException
+     * @throws ServerErrorHttpException
+     */
+    public static function create($userId, $value)
+    {
+        $method = Method::findOne(['value' => $value, 'user_id' => $userId]);
+
+        if ($method === null) {
+            $method = new Method;
```
forevermatt: Very minor issue, but parentheses are recommended even when instantiating a class that doesn't take any arguments.
```suggestion
            $method = new Method();
```
I think the main place I learned this was the (pending) [PSR-12](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#4-classes-properties-and-methods). Even though it hasn't been finalized, it seemed worth learning from it.

application/features/bootstrap/fakes/FakeEmailer.php
```diff
@@ -38,28 +38,32 @@ public function forgetFakeEmailsSent()
      *
      * @param string $messageType The type of message.
      * @param User $user The User in question.
+     * @param string $address Email address to find. If provided, overrides $user->email.
      * @return array[]
      */
     public function getFakeEmailsOfTypeSentToUser(
```
forevermatt: Rather than adding a 3rd parameter (which, when present, makes the 2nd parameter unused), what would you think about changing this method to be this?
```php
public function getFakeEmailsOfTypeSentToAddress(string $messageType, string $emailAddress)
```

application/common/models/Method.php
```diff
@@ -177,4 +174,33 @@ public static function deleteExpiredUnverifiedMethods()
...
     }
+
+    /**
+     * @param integer $userId
+     * @param string $value
+     * @return Method
+     * @throws ConflictHttpException
+     * @throws ServerErrorHttpException
+     */
+    public static function create($userId, $value)
```
forevermatt: This may be more of a change that you want to make at the moment, but it seems like this function's name doesn't quite line up with what it does. Do you think it would be worth renaming this to `findOrCreate(...)`?

application/common/components/Emailer.php
```diff
@@ -190,16 +192,27 @@ public function getSiteStatus()
...
         if ( ! empty($this->subjects[$messageType]) && strpos($this->subjects[$messageType], '%') !== false) {
             return sprintf($this->subjects[$messageType], $this->otherDataForEmails['idpDisplayName'] ?? '');
         }
         return $this->subjects[$messageType] ?? null;
     }
-    
-    protected function getViewForMessage(string $messageType, string $format)
+
+    /**
+     * Retrieve the view text for the given message type and format. The view is read from a
```
forevermatt: This may just be a matter of clarification, but this method actually gets the path to the view file for the given message type and format. That path is later passed to a render() function, which goes and gets that view file and uses it in generating the content for the email.

It may just be me, but the way I understood your method description made it sound like it actually got the content of the message for the given message type an format.

application/features/bootstrap/EmailContext.php
```diff
@@ -905,4 +911,63 @@ public function iSeeThatTheSecondUserHasReceivedAGetBackupCodesEmail()
...
+        $this->createMethod(self::METHOD_EMAIL_ADDRESS);
+    }
+
+    /**
+     * @Then a Method Verify email is sent to that method
+     */
+    public function aMethodVerifyEmailIsSentToThatMethod()
+    {
+        $matchingFakeEmails = $this->fakeEmailer->getFakeEmailsOfTypeSentToUser(
+            EmailLog::MESSAGE_TYPE_METHOD_VERIFY, self::METHOD_EMAIL_ADDRESS);
```
forevermatt: Again a very minor issue, but some PSR-2 checker would probably complain about this since the arguments to the function were long enough to wrap, but weren't each on their own separate line.
```suggestion
            EmailLog::MESSAGE_TYPE_METHOD_VERIFY,
            self::METHOD_EMAIL_ADDRESS
        );
```

PR #145, opened 2018-11-30
Add "do not disclose" flag to user model and API

Schparky:
- Added `do_not_disclose` to user table
- Added `do_not_disclose` as required property in user object in API
- Refactored RAML file to be more easily read by human eyes

### DIFF
--- a/api.raml
+++ b/api.raml
@@ -58,6 +58,7 @@ types:
       last_login_utc?: string
       manager_email?: string
       spouse_email?: string
+      do_not_disclose: boolean
       mfa?:
         type: object
         properties:
@@ -89,6 +90,7 @@ types:
         "active": "yes",
         "locked": "no",
         "last_login_utc": "2017-05-24 14:04:51 UTC",
+        "do_not_disclose": false,
         "mfa": {
           "prompt": "yes",
           "nag": "no",
@@ -132,6 +134,7 @@ types:
         enum: [ yes, no ]
       manager_email?: string
       spouse_email?: string
+      do_not_disclose: boolean
     example: |
       {
         "employee_id": "12345",
@@ -139,6 +142,7 @@ types:
         "last_name": "Smith",
         "username": "john_smith",
         "email": "john_smith@example.com",
+        "do_not_disclose": false
       }
   UserUpdate:
     type: object
@@ -154,6 +158,7 @@ types:
         enum: [ yes, no ]
       manager_email?: string
       spouse_email?: string
+      do_not_disclose?: boolean
     example: |
       {
         "active": "no"
--- a/api.raml
+++ b/api.raml
@@ -40,8 +40,8 @@ types:
       }
   ReturnedUser:
     description: >
-      Information on user record. Sub-objects mfa and method are included
-      except in a user list.
+      Information on user record. Password is not included if a password
+      has not been assigned to the user.
     type: object
     properties:
       uuid?: string
@@ -59,7 +59,7 @@ types:
       manager_email?: string
       spouse_email?: string
       do_not_disclose: boolean
-      mfa?:
+      mfa:
         type: object
         properties:
           prompt:
@@ -67,7 +67,7 @@ types:
           nag:
             enum: [ yes, no ]
           options: Mfa[]
-      method?:
+      method:
         type: object
         properties:
           nag:
--- a/api.raml
+++ b/api.raml
@@ -14,488 +14,288 @@ securitySchemes:
 securedBy: [ AuthzBearerToken ]
 
 types:
-  Authentication: |
+  Authentication:
+    type: object
+    properties:
+      username: string
+      password: string
+    example: |
       {
-        "$schema": "http://json-schema.org/draft-03/schema",
-        "properties": {
-          "username": {
-            "required": true,
-            "type": "string"
-          },
-          "password": {
-            "required": true,
-            "type": "string"
-          }
-        },
-        "type": "object"
+        "username": "john_smith",
+        "password": "correcthorsebatterystaple"
       }
-  Error: |
+  Error:
+    type: object
+    properties:
+      name: string
+      message: string
+      code: integer
+      status: integer
+    example: |
       {
-        "$schema": "http://json-schema.org/draft-03/schema",
-        "properties": {
-          "name": {
-            "required": true,
-            "type": "string"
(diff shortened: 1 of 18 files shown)


api.raml
```diff
@@ -14,488 +14,293 @@ securitySchemes:
...
+      email?: string
+      active?:
+        enum: [ yes, no ]
+      locked?:
+        enum: [ yes, no ]
+      last_login_utc?: string
+      manager_email?: string
+      spouse_email?: string
+      do_not_disclose: boolean
+      mfa?:
```
forevermatt: `mfa` was marked as required in the previous version of the RAML file. Was it intentionally made optional?

application/console/migrations/m181128_213015_add_do_not_disclose_column_to_user_table.php
```diff
@@ -0,0 +1,25 @@
...
+ * Handles adding do_not_disclose to table `user`.
+ */
+class m181128_213015_add_do_not_disclose_column_to_user_table extends Migration
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function safeUp()
+    {
+        $this->addColumn('user', 'do_not_disclose', 'boolean not null default false');
```
forevermatt: It seems like we have some data duplication between idp-pw-api and idp-id-broker. Why do they both need a `do_not_disclose` field?

application/frontend/config/main.php
```diff
@@ -20,6 +20,7 @@
         ],
         // http://www.yiiframework.com/doc-2.0/guide-runtime-requests.html
         'request' => [
+            'enableCookieValidation' => false,
```
forevermatt: I'm curious... why the cookie validation change?

application/frontend/controllers/UserController.php
```diff
@@ -53,6 +58,11 @@ public function actionUpdate(string $employeeId)
 
         $user->attributes = Yii::$app->request->getBodyParams();
 
+        /*
+         * Received as boolean, stored as tinyint(1)
+         */
+        $user->do_not_disclose = (int)$user->do_not_disclose;
```
forevermatt: Is this type coercion one we have to do manually? I ask because I don't remember this being an issue before with boolean columns in the database.

api.raml
Schparky: Great question. After revisiting this, I found that it is an artifact of the fact that the previous version defined a separate `ReturnedUser` and `UserList`. I combined them and used `ReturnedUser[]` in place of `UserList`. It seems that the definition of `UserList` was incorrect in that it omitted `Mfa` and `Method`. So to be correct I need to make them required in the new version.
As for `password`, it didn't change. It is not always included, so it is correct to marked as optional.

application/console/migrations/m181128_213015_add_do_not_disclose_column_to_user_table.php
Schparky: It was a design decision influenced by the way idp-pw-api keeps a copy of user data. I'm open to alternate ideas if you would like to discuss.

application/frontend/config/main.php
Schparky: Yii2 was updated from 2.0.13 to 2.0.15. I don't know the background of why this was added, other than 2.0.15 has some security fixes.

application/frontend/controllers/UserController.php
Schparky: Unfortunately, yes. The type validation in Yii doesn't recognize `false` as an integer. Interestingly, `true` **is** recognized as an integer. I found the reason has to do with the way PHP casts `bool` to `string`. `true` becomes '1' and `false` becomes ''
I did notice that some other database columns use enums such as 'yes', 'no'. Perhaps this solution circumvents the problem.

application/frontend/config/main.php
forevermatt: It looks like this change may have been the result of partially-copied code from another repo. IIRC, once Yii started requiring `enableCookieValidation` to be set, we added a way for us to provide a cookie validation key via an env. var.

For example, see the various occurrences of "cookieValidation" (case-insensitive) in this commit on idp-id-sync:
https://github.com/silinternational/idp-id-sync/blob/3732472b7b3b2bac43444134b3cc953b95ad7ac2/application/frontend/config/main.php

I'd recommend going ahead and adding...
```php
$cookieValidationKey = Env::get('COOKIE_VALIDATION_KEY');
```
... up above this array in this file, then changing this line per the following (as part of dealing with the upgrade to Yii 2.0.15):
```suggestion
            'cookieValidationKey' => $cookieValidationKey,
            'enableCookieValidation' => !empty($cookieValidationKey),
```

application/console/migrations/m181128_213015_add_do_not_disclose_column_to_user_table.php
forevermatt: I'll go ahead and look at that PR (and you may have addressed this already), but we'll probably need to make sure that either (1) pw-api gets the latest value for `do_not_disturb` from id-broker every time it uses the value or (2) id-broker sends `do_not_disturb` changes to pw-api every time it changes there in id-broker. I believe we will eventually allow this field to be set via id-sync, based on data from the HR system, so it could change it id-broker by a means other than pw-api.

forevermatt: I'd like to discuss this with @fillup to make sure I'm not mixing things up here (re: changes to `do_not_disturb` in id-broker that would leave it out-of-sync with pw-api).

forevermatt: Okay, @Schparky. @fillup talked me through some of the design decisions and trade-offs, and it's fine for you to go proceed with both pw-api and id-broker having a `do_not_disclose` field. The pw-api's copy seems to work effectively as a cache, since it's updated from id-broker on login. It also looks like id-broker's value will very likely only be updated through pw-api... the integrations with HR systems to add another field and train people how to use it added enough complexity that it's better for us to just keep that setting entirely within this IdP-in-a-Box system.

Schparky: Good. Thanks for taking a good amount of time to work through it. Much appreciated.

PR #147, opened 2018-12-05
Change do_not_disclose column to enum

Schparky:
As suggested by @wcjr and @fillup, to be consistent and more error-resistant, change the `do_not_disclose` column and property from int/bool to string/enum.


api.raml
```diff
@@ -90,7 +91,7 @@ types:
         "active": "yes",
         "locked": "no",
         "last_login_utc": "2017-05-24 14:04:51 UTC",
-        "do_not_disclose": false,
+        "do_not_disclose": "no",
```
wcjr: :-D  oh man, this double-negative is a little hard to read...it might've been better once we realized that yes/no stuff to name this `disclose` or `sensitive`... not a biggie I don't suppose at this point though.

wcjr: I think `disclose` is better but let me check the thesaurus and see if there's something better.

wcjr: unveil :-D

wcjr: is *protect* any better than disclose?

Schparky: I would prefer a word with positive logic (antonym of disclose) so that the true/false sense doesn't change. In other words, I'd still like to have "no" be the default.

Schparky: How about _conceal_?

wcjr: oh yeah, I like that.

fillup: sure, that works for me

forevermatt: It looks like you've come to a consensus already, so feel free to reject this, but would "hide" be any better than "conceal"? It just seems like a word we'd be more likely to use in a normal conversation.

PR #148, opened 2018-12-06
Add recovery method and MFA review flags

Schparky:
- changed `nag` property in `mfa` sub-object of userResponse to `add`
- added `add` property in `method` sub-object of userResponse
- added `review` property in both `mfa` and `method` sub-objects of userResponse


application/behat.yml
```diff
@@ -5,11 +5,11 @@ default:
                 - "%paths.base%/features/authentication.feature"
                 - "%paths.base%/features/password.feature"
                 - "%paths.base%/features/user.feature"
-            contexts: [ FeatureContext ]
+            contexts: [ FeatureContext, Sil\SilIdBroker\Behat\Context\UnitTestsContext ]
```
forevermatt: Interesting, I don't remember seeing multiple Contexts listed before. in the past, I've simply had the more specific context (e.g. UnitTestsContext) be a subclass of the broader contexts (FeatureContext). Do you happen to know of any advantages of listed multiple contexts vs. subclassing?

application/common/models/User.php
```diff
@@ -426,13 +435,42 @@ public function getDisplayName()
...
+     * determine which "nag" to present to the user.
      */
-    public function isTimeToNagForMfa()
+    public function getNagState()
     {
-        return strtotime($this->nag_for_mfa_after) < time();
+        /*
+         * Don't recalculate in case the date has changed since the last calculation.
+         */
+        if ($this->nagState !== null) return $this->nagState;
```
forevermatt: Per PSR-2, you'll want to switch this to enclose the if-block in curly braces:
```suggestion
        if ($this->nagState !== null) {
            return $this->nagState;
        }
```

application/common/models/User.php
```diff
@@ -426,13 +435,42 @@ public function getDisplayName()
...
+
+        $now = time();
+
+        if (strtotime($this->nag_for_mfa_after) < $now) {
```
forevermatt: As a side note, [strtotime](https://php.net/strtotime) returns `false` if it fails to understand the date/time string, so if `$this->nag_for_mfa_after` isn't understandable, this condition will evaluate to true and the user will be nagged.

If that's a problem, it may be worth extracting this logic to a new method that intelligently handles invalid date/time strings somehow.

application/common/models/User.php
```diff
@@ -426,13 +435,42 @@ public function getDisplayName()
...
+            if (count($this->getVerifiedMfaOptions()) === 0) {
+                $this->nagState = self::NAG_ADD_MFA;
+            } else {
+                $this->nagState = self::NAG_REVIEW_MFA;
```
forevermatt: So it looks like we'll be using the `nag_for_mfa_after` database field for two related but not quite identical purposes:
1) How long to wait before nagging the user to turn on MFA (by adding one), and
2) How long to wait (once they are using MFA) to have them review their MFA settings.

Is that right?

Those seem related enough that I don't see a problem with that (as long as the time-between-reviews can be different than the time-between-nags-to-add-mfa), but I wanted to make sure I understood it correctly and that we're doing that intentionally.

application/common/models/User.php
```diff
@@ -426,13 +435,42 @@ public function getDisplayName()
...
 
     /**
-     * Is it time to nag the user to add an MFA?
-     *
-     * @return bool
+     * Based on current time and presence of MFA and Method options,
+     * determine which "nag" to present to the user.
      */
-    public function isTimeToNagForMfa()
+    public function getNagState()
```
forevermatt: I find the logic of this method a little hard to follow. What is our order of priority for which nag (when applicable) should happen?

I wonder if there's a way to use an array of closures keyed on NAG_* constants, where we could just use the first one that returns true. Something like the following (though my example code's syntax might not be valid):
```php
public function getNagState()
{
    if ($this->nagState !== null) {
        return $this->nagState;
    }
    
    $possibleNags = [
        self::NAG_ADD_MFA => function isTimeToNagToAddMfa() { ... },
        self::NAG_REVIEW_MFA => function isTimeToNagToReviewMfa() { ... },
        self::NAG_ADD_METHOD => function isTimeToNagToAddMethod() { ... },
        self::NAG_REVIEW_METHOD => function isTimeToNagToReviewMethod() { ... },
        self::NAG_NONE => true,
    ];
    
    foreach ($possibleNags as $nagType => $isTime) {
        if ($isTime()) {
            $this->nagState = $nagType;
            return $this->nagState;
        }
    }
}
```

application/common/models/User.php
```diff
@@ -442,11 +480,15 @@ public function getMfaFields()
     {
         return [
             'prompt'  => $this->isPromptForMfa() ? 'yes' : 'no',
-            'nag'     => (! $this->isPromptForMfa() && $this->isTimeToNagForMfa()) ? 'yes' : 'no',
+            'add'     => $this->getNagState() == self::NAG_ADD_MFA ? 'yes' : 'no',
```
forevermatt: Are we going to have logic here in ID Broker to ensure that, if `prompt` is `yes` then no nags will happen? It seems like it would be good to avoid doing both, if we can, but I don't see anything here that would prevent them both from happening at the same time (the way we did in the old code).

application/common/models/User.php
Schparky: I don't see it as a problem since `nag_for_mfa_after` is coming from the database. Will consider this potential issue in the future.

application/common/models/User.php
Schparky: Yes, that is correct. That's why I changed the `nag` json property to `add` in order to keep semantics clear. "nag" is a general term for user reminder, and can either be an "add" or "review"

application/common/models/User.php
```diff
@@ -753,19 +799,16 @@ public function isPromptForMfa(): bool
...
-     * Update the first nag_for_*_after date that has passed
+     * Update the nag_for_*_after date that corresponds to the current nag state
      */
     public function updateNagDates()
     {
-        // save current time to avoid race condition
-        $now = time();
-
-        if (strtotime($this->nag_for_mfa_after) < $now) {
+        if ($this->getNagState() === self::NAG_ADD_MFA || $this->getNagState() === self::NAG_REVIEW_MFA) {
```
forevermatt: I think we'll want the interval between nags-to-set-up-MFA to be different than the interval between nags-to-review-your-MFA. For example, if I haven't set up MFA, I could understand being nagged every couple of months. However, if I have set up MFA, I wouldn't want to be nagged about reviewing it more often than every 6 - 12 months.

application/common/models/User.php
Schparky: Good idea. I'll see if it works.

application/common/models/User.php
Schparky: I looked at the consumer code you sent, and it was clear that `prompt` was the priority. If `prompt` is `yes`, then `nag` is ignored. Didn't make sense to duplicate the logic here and prevent the consumer from implementing the logic as best fits a future scenario. I even considered making `getNagState` obsolete by simply providing all appropriate nags and leaving the priority up to the consumer, though I didn't go that far since it wasn't the original plan.

application/features/analytics.feature
```diff
@@ -3,9 +3,11 @@ Feature: User
   In order to report about mfa usage
   I need to be able to count active users with diffent kinds of verified mfa records
 
+  Background:
+    Given the user store is empty
```
forevermatt: This is somewhat subjective, but it seems to be more helpful in general to keep Behat statements as customer-friendly as reasonably possible. For example, rather than saying "the user store is empty" it might be better to say "there are no users" (or possibly "there are no users in the database").

The thought behind it is that the *.feature files should contain language that the customer / product owner can read and understand to bring consensus on what qualifies as "done". Since this are tests for components without our own system, that customer interaction is even less likely to happen, but keeping that goal still seems valuable.

Schparky: OK, thanks. I'll write feature statements with that in mind.

application/features/analytics.feature
```diff
@@ -3,9 +3,11 @@ Feature: User
...
+
   Scenario: Get a count of active users with a verified mfa
-    Given that no mfas or users exist
-      And I create a new user
+      Given I create a new user
```
forevermatt: As another side note about Behat/Gherkin, if you can find ways to make the `Given`'s declarative rather than imperative statements (e.g. "1 user exists who has ____" rather than "I create a new user"), it seems to result in more readable tests. The `Given`'s are simply setting up the state of the system before it's tested, whereas the `When` is where we want a specific action to be taken.

That said, I realize you're simply adding to and/or revising an existing set of tests, and it certainly makes sense to use existing steps rather than write new steps unnecessarily. Feel free to proceed along your existing route; I just wanted to share some things I've learned, for future code.

Schparky: Good thoughts. Thanks for sharing!

application/features/authentication.feature
```diff
@@ -177,6 +177,37 @@ Feature: Authentication
...
+      | password  | govols!!!   |
+    When I request "/authentication" be created
+    Then the following data is returned:
+      | property      | value                 |
+      | employee_id   | 123                   |
+      | method.add    | no                    |
+      | method.review | no                    |
+      | mfa.add       | yes                   |
+      | mfa.review    | no                    |
+    When I request "/authentication" be created
```
forevermatt: The presence of multiple `When`'s here is an indicator that this single scenario may be trying to test too many things. A good rule of thumb can be that [a scenario should have only one When step](https://stackoverflow.com/a/19401196).

I don't know whether it's worth the time, but you might consider writing separate, simpler scenarios for checking the whether the user will be nagged for various scenarios. Example:
```gherkin
  Scenario: Nagging a user to set up MFA
    Given I have provided valid credentials
      And I have not been nagged to set up MFA
      And I have not been nagged to set up a recovery method
    When I log in
    Then I should only be nagged to set up MFA

  Scenario: Nagging a user to set up a recovery method
    Given I have provided valid credentials
      And I have been nagged to set up MFA recently
      But I have not been nagged to set up a recovery method
    When I log in
    Then I should only be nagged to set up a recovery method

  Scenario: Not re-nagging a user right away
    Given I have provided valid credentials
      And I have been nagged to set up MFA recently
      And I have been nagged to set up a recovery method recently
    When I log in
    Then I should not be nagged to set up MFA
      And I should not be nagged to set up a recovery method
```

application/behat.yml
Schparky: Also, I recall a situation where it didn't make sense (or was impossible) to combine contexts into a class hierarchy. Though I can't remember the specifics, it had something to do with PHP not having multiple inheritance. If I remember or run across it again, I'll let you know.

application/common/models/User.php
```diff
@@ -21,11 +21,20 @@ class User extends UserBase
...
+    const NAG_ADD_METHOD    = 'add_method';
+    const NAG_REVIEW_METHOD = 'review_method';
+
     public $password;
 
     /** @var Ldap */
     private $ldap;
 
+    /** @var string */
+    public $nagState;
```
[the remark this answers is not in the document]
Schparky: Actually, it was easy. I just added a `resetNagState` method.

application/features/bootstrap/UnitTestsContext.php
```diff
@@ -0,0 +1,170 @@
...
+     * @Then I update the nag dates
+     */
+    public function iUpdateTheNagDates()
+    {
+        $this->tempUser->updateNagDates();
+
+        /*
+         * Force an update the next time getNagState() is called
+         */
+        $this->tempUser->nagState = null;
```
forevermatt: Is this something that it would be good to do every time the nag dates are updated (and thus, that it would potentially be worth moving into the `updateNagDates()` method)?

application/features/user.feature
```diff
@@ -401,6 +404,42 @@ Feature: User
     Then the response status code should be 200
       And I should receive 2 users
 
+  Scenario: Get list of verified methods for a user
+    Given A user with 1 verified methods, 1 unverified method, 0 verified mfas, and 0 unverified mfas
```
forevermatt:
```suggestion
    Given A user with 1 verified method, 1 unverified method, 0 verified mfas, and 0 unverified mfas
```

application/features/bootstrap/UnitTestsContext.php
Schparky: Perhaps. It's a question of how it might be used in the future. As of now, it wouldn't make a difference. Since `authenticatedUser` is cloned prior to calling `updateNagDates`, it would be safe to reset the `nagState` in `updateNagdates`. Logic seems to be in favor of doing as you suggest.

application/common/models/User.php
Schparky: I handled this while refactoring getNagState

application/features/authentication.feature
Schparky: Good question. Is it worth the time to rewrite this? IMHO, no, but worth considering.

application/common/models/User.php
Schparky: Done.

application/common/models/User.php
Schparky: Yep. Worked perfectly.

application/features/bootstrap/UnitTestsContext.php
Schparky: Done.

application/behat.yml
forevermatt: I think it's fine to have multiple Contexts, it was just something I hadn't done before. It sounds like in this case it was the better choice.

application/common/models/User.php
forevermatt: Okay, great. Glad you did it this way, then.

PR #149, opened 2018-12-12
Added "new user invite" authentication

Schparky:
- new table in database
- new `invite` property on `/authentication`
- added invite code to invite email template


application/common/config/main.php
```diff
@@ -78,7 +78,7 @@
                 'supportName' => Env::get('SUPPORT_NAME', 'support'),
             ],
             
-            'sendInviteEmails' => Env::get('SEND_INVITE_EMAILS', false),
+            'sendInviteEmails' => Env::get('SEND_INVITE_EMAILS', true),
```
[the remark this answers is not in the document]
Schparky: Seemed appropriate for the new "new user" work flow. I'll check with @fillup to confirm.

application/common/models/User.php
```diff
@@ -316,6 +333,7 @@ public function getAttributesForEmail()
             'isMfaEnabled' => count($this->mfas) > 0 ? true : false,
             'mfaOptions' => $this->getVerifiedMfaOptions(),
             'numRemainingCodes' => $this->countMfaBackupCodes(),
+            'inviteCode' => Invite::getInviteCode($this->id)->uuid,
```
forevermatt: It looks like the call to `Invite::getInviteCode(...)` will end up creating an invite code for every user that an email is ever sent to, even when they already have a password and we have no desire to create an invite code for them. That seems worth changing.

application/common/models/Invite.php
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Return an existing, non-expired invite code, or create a new object and
+     * return the new code.
+     *
+     * @param int $userId
+     * @return Invite
+     * @throws \Exception
+     */
+    public static function getInviteCode(int $userId): Invite
```
forevermatt: This method's name doesn't quite seem to line up with what it actually does: it says it will get an invite code, but it actually returns an Invite instance. I'm not sure which of those you'll end up wanting a function for (possibly both), and that will probably guide which thing you change here.

application/common/config/main.php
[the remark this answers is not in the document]
Schparky: He confirmed via slack.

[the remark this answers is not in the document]
forevermatt: Okay, thanks.

application/common/models/User.php
Schparky: Oh, yikes. I'll work on that.

Schparky: Done.

application/common/models/Invite.php
Schparky: Done.

application/console/migrations/m181207_144943_create_invite_table.php
```diff
@@ -0,0 +1,38 @@
...
+     * {@inheritdoc}
+     */
+    public function safeUp()
+    {
+        $this->createTable(
+            'invite',
+            [
+                'id' => 'pk',
+                'user_id' => 'int(11) not null',
+                'uuid' => 'char(36) CHARACTER SET ascii COLLATE ascii_general_ci not null',
```
fillup: Does the ascii thing matter?

application/console/migrations/m181207_144943_create_invite_table.php
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function safeUp()
+    {
+        $this->createTable(
+            'invite',
+            [
+                'id' => 'pk',
+                'user_id' => 'int(11) not null',
```
fillup: Do we want/need to allow users to have multiple invites?

application/console/migrations/m181207_144943_create_invite_table.php
Schparky: Probably not. I was thinking that it would reduce storage space, but I recognize it's trivial.

application/console/migrations/m181207_144943_create_invite_table.php
Schparky: No. Is there an easy way to enforce 1:1?

Schparky: After further thought, I realize I could make `user_id` unique. However, to keep moving, I will not change anything here. The current business logic should prevent multiple invites for any user. Actually, to be precise, it's not possible for multiple _non-expired_ invites per user. Presently, there is no garbage collection. I'll make a task card to write a cron script to clean up expired records.

PR #152, opened 2018-12-18
Restart verification for expired, unverified methods

Schparky:
For a verify request on an expired, unverified method, do the following:
- validate the provided code
- set the attempts counter to 1
- create a new code
- extend the expiration
- send an email with the new code
- respond with a 410 status code


Schparky:
@forevermatt Could you take a look? @wcjr deferred the algorithm review on this one.

PR #153, opened 2019-01-02
Renew expired invite

Schparky:
In response to `/authentication` with an `invite` code that has expired, respond with a 410 and renew the invite with a new code. Also send an email to the user with the new code.

### DIFF
--- a/application/common/models/Authentication.php
+++ b/application/common/models/Authentication.php
@@ -67,7 +67,7 @@ class Authentication
      * Attempt an authentication by new user invite.
      *
      * @param string $invite New user invite code. If not blank, username and password are ignored.
-     * @throws \Exception
+     * @throws HttpException
      */
     protected function authenticateByInvite($invite)
     {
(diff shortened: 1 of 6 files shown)


application/common/models/Authentication.php
```diff
@@ -74,9 +77,15 @@ protected function authenticateByInvite($invite)
             $this->errors['invite'] = ['Invalid code.'];
             return;
         }
+
         if ( ! $invite->isValidCode()) {
```
wcjr: it looks like an invalid code that is NOT expired will generate a new code and email, am I reading that right?

Schparky: I was afraid of that being misleading. The previous `if` ensures that the code is a real code. `isValidCode` only checks expiration.

wcjr: would changing the name to `isExpired` avoid the confusion?

Schparky: done

PR #156, opened 2019-01-10
Manager rescue MFA

Schparky:
Added a new MFA type `manager` that is similar to `backupcode` except for the following:
- only one code is created
- the code is immediately emailed to the user's manager

Also:
- deleted composer.json and composer.lock in root directory as these were not being used
- added MFA feature scenarios to cover more use cases 
- adjusted indention in feature files


forevermatt:
I'm a little concerned about this design. Adding a "Manager rescue MFA" option means that a manager can successfully impersonate one of their employees on any system that IdP is connected to, even if that employee has 2SV enabled. @fillup - is that intentional?

application/common/components/Emailer.php
```diff
@@ -30,6 +30,8 @@ class Emailer extends Component
     const SUBJECT_MFA_OPTION_REMOVED_DEFAULT = 'A 2-Step Verification option was removed from your %s account';
     const SUBJECT_MFA_ENABLED_DEFAULT = '2-Step Verification was enabled on your %s account';
     const SUBJECT_MFA_DISABLED_DEFAULT = '2-Step Verification was disabled on your %s account';
+    const SUBJECT_MFA_MANAGER_DEFAULT = 'Someone has sent you a login code for their %s account';
```
fillup: Can we put the individual's name in here instead of `Someone`? That'll help the manager recognize it as importantI think

Schparky: Yes, I'll have to write some code to expand the ability to put variables into subject lines. I was considering it, but hadn't bit the bullet.

PR #157, opened 2019-01-22
add method options to UserResponse; created to MethodResponse

Schparky:
- added `method.options` to UserResponse schema, used on all API responses that return user data
- added `created` to MethodResponse schema, used on all API responses that return method data
- changed `uid` to `id` in MethodResponse API doc, to agree with code functionality


application/common/models/User.php
```diff
@@ -570,6 +570,7 @@ public function getMethodFields()
         return [
             'add' => $this->getNagState() == self::NAG_ADD_METHOD ? 'yes' : 'no',
             'review' => $this->getNagState() == self::NAG_REVIEW_METHOD ? 'yes' : 'no',
+            'options' => $this->methods,
```
fillup: We discussed not including the options/methods if they don't need to be reviewed, do you want to update that before we approve/merge this or come back to it?

Schparky: Pushed the update just now.

PR #158, opened 2019-01-24
Remove SQL CASCADE foreign key constraints

Schparky:
- Changed `email_log` and `password` constraints from `CASCADE` to `NO ACTION`
- Added PHP deletes for `email_log` and `password` relationships
- Added `EmailLog::deleteAll` in test code where `User::deleteAll` is used

**Note 1**: `NO ACTION` is equivalent to `RESTRICT` in MySQL, effectively disabling `delete` for *parent records, without first deleting child records.*
**Note 2**: The original problem with `CASCADE` was because *of the placement of the foreign key constraint on the `user` table. The constraints on other tables could be of type `CASCADE`.*


forevermatt:
I also wonder (re: Note 1) if keeping the "on delete" actions set to prevent deletion of the parent record if child records still exist would actually be helpful. That would force us to intentionally delete the child records if a parent record is being deleted, which we probably want to do anyway (using individual `delete()` calls) so that any business logic we have in `beforeDelete()` and/or `afterDelete()` on the child records is applied rather than bypassed. Thoughts?

Schparky:
Regarding Note 2, although it's true that there are foreign keys on both tables (`user` and `password`), there is a CONSTRAINT only on `user`. Based on my current understanding and recent tests and troubleshooting, a `CASCADE` constraint will delete records on the table where the constraint is defined. In this case, the `user` table. 

Likewise, with `NO ACTION` constraints, a delete is prevented on the table that contains the constraint. It surprised me, and I spent a good while trying to figure out why the child `delete` in `User::beforeDelete` was failing. I don't see any way to make a `RESTRICT` or `NO ACTION` constraint work, unless I simply made a mistake in my troubleshooting (which is all too possible). It may work to delete the constraints altogether. I hadn't really considered that too much.

I just had another thought. Maybe we could use `NO ACTION` and do the child delete in `afterDelete` instead of `beforeDelete`. I might play with that today.

Schparky:
After further study, I found I was wrong about `NO ACTION` preventing deletion of child records. I misdiagnosed the problem. In reality it was something completely different (deletion of `mfa` records creates a new `email_log` record, so `mfa` deletes must occur before `email_log` deletes). So I changed the constraints to `NO ACTION` as @forevermatt originally had done, instead of `SET NULL`. I edited the pull request summary comment accordingly.

application/common/models/User.php
```diff
@@ -93,6 +122,22 @@ public function beforeDelete()
             }
         }
 
+        /*
+         * Delete email logs last in case other deletions trigger new emails
```
forevermatt: So, _does_ deleting something else trigger an email? For example, if deleting an mfa record sends an email to the user, is that a problem when we're doing that in order to delete that user?

I think we're only going to delete users if they've been inactive for over a year, but I do wonder about the repercussions of emailing a user who is about to be purged from our system...

Schparky: Nice catch. So, to extrapolate this, would it be fair to say that no email should be sent to inactive users?

Schparky: Since this is a separate, preexisting issue, I'm going to proceed with merging this PR and create a separate issue/PR for the problem of sending an email to an about-to-be-deleted user.

Schparky:
For the record, a change was made to block sending a message to an inactive user: https://github.com/silinternational/idp-id-broker/commit/eaca7eeaceb38bf31c376d62d405502064d450cb

PR #161, opened 2019-02-08
set default mfa label based on its type

Schparky:


wcjr:
@fillup and I had thought the default label for backupcodes probably should change since that one is way more static.  This does NOT address that too does it?

Schparky:
This does address backupcode labels. It uses the existing `getReadableType()` which uses "Printable Codes" for mfa type `backupcode`.

PR #165, opened 2019-02-26
Restrict POST /mfa to only allow one of each type of MFA

Schparky:
Returns http 409 if attempting to create a second of the requested type


Schparky:
> any changes to the api docs required?

Yes, thank you!

PR #166, opened 2019-03-04
cc personal_email on new user invite

Schparky:


Schparky:
> One idea for a slight improvement if you work in the Emailer.php file more is to use constants for all the keys for the $data array ('toAddress', 'ccAddress', ...) and then, as well, the code that sends a $data param to an Emailer.php method, could build it up using those constants.

Thanks, good tip. I also thought of breaking up the `email()` method into individual calls to set the properties before actually sending the email. Something like:
```php
$emailer
    ->setTo($toAddress)
    ->setSubject($subject)
    ->setHtmlBody($body)
    ->send();
```
Pretty much a copy of the Yii message interface. Thoughts?

Baggerone:
The idea of breaking up the email() method doesn't really resonate with me.  I'm not sure what you would gain.  It would make sense if there are situations when you want to just set some but not all of the properties.

Schparky:
The 'cc' email address is not always needed. It also encapsulates each property with its own method rather than a string constant. Might also be helpful if/when we need to allow an array of email addresses, but that's just speculation. But for now, I agree that there's not much benefit.

PR #171, opened 2019-03-15
Create contingent user

Schparky:
Allows for creation of a "contingent" user -- a user that is in the onboarding process, needs access, but does not yet have a primary email address.
- Adds concept of user record expiration, set to 4 weeks by default, and not defined for normal user creation
- `personal_email` is provided everywhere that `email` is needed if `email` is `null`: outgoing email message "To" header, email template variable, and API responses
- `email` can only be set to `null` on initial creation of a user record, and only if `personal_email` is not `null`

### DIFF
--- a/application/common/config/main.php
+++ b/application/common/config/main.php
@@ -219,5 +219,6 @@ return [
             Env::getArrayFromPrefix('METHOD_')
         ),
         'mfaLifetime' => Env::get('MFA_LIFETIME', '+2 hours'),
+        'contingentUserDuration' => Env::get('CONTINGENT_USER_DURATION', '+4 weeks'),
     ],
 ];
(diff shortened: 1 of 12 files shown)


application/frontend/controllers/AuthenticationController.php
```diff
@@ -19,19 +19,37 @@ public function actionCreate(): User
     {
         $migratePasswords = Yii::$app->params['migratePasswordsFromLdap'];
 
+        $username = (string)Yii::$app->request->getBodyParam('username');
```
wcjr: I like that change, much cleaner

application/common/config/main.php
```diff
@@ -219,5 +219,6 @@
             Env::getArrayFromPrefix('METHOD_')
         ),
         'mfaLifetime' => Env::get('MFA_LIFETIME', '+2 hours'),
+        'contingentUserDuration' => Env::get('CONTINGENT_USER_DURATION', '+4 weeks'),
```
wcjr: thank you for giving a default!  Do you mention this new env var in the docs or README?

Schparky: If `local.env.dist` counts as a doc, yes. I just checked and `README.md` doesn't mention env vars.

PR #172, opened 2019-04-04
Add new endpoint to assess whether a new password is acceptable

Schparky:
This new endpoint (`PUT /user/{id}/password/assess`) does the same checks as for updating a password (`PUT /user/{id}/password`) but does not save the new password.

### DIFF
--- a/api.raml
+++ b/api.raml
@@ -443,17 +443,6 @@ types:
               type: UserResponse
           204:
             description: No user record was found with that `employee_id`.
-          409:
-            description: A recently-used password was given.
-            body:
-              type: Error
-              example: |
-                {
-                  "name": "Conflict",
-                  "message": "May not be reused yet.",
-                  "code": 1542395933,
-                  "status": 409
-                }
           422:
             description: >
               The given password does not meet some requirement (such as if an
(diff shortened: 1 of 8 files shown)


api.raml
```diff
@@ -443,17 +443,6 @@ types:
               type: UserResponse
           204:
             description: No user record was found with that `employee_id`.
-          409:
```
wcjr: I'm surprised to see this removed...am I missing something?

api.raml
```diff
@@ -464,6 +453,34 @@ types:
...
+              }
+          responses:
+            204:
+              description: >
+                The `password` is valid for the user with the given `employee_id`.
+            404:
+              description: No user found with given `employee_id`.
+              body:
+                type: Error
+            422:
```
wcjr: `409`?

api.raml
Schparky: My mistake. At one point I had considered changing it to 422 like all other model validations. I'll correct this.

application/common/models/User.php
```diff
@@ -914,4 +914,19 @@ public function afterFind()
...
+
+    public function assessPassword(string $newPassword): bool
+    {
+        $password = new Password();
+
+        $password->user_id = $this->id;
+        $password->password = $newPassword;
+
+        if (! $password->validate()) {
+            $this->addErrors($password->getErrors());
```
forevermatt: Are the errors returned by `getErrors()` associated with specific attribute names? I didn't know if taking a `Password` record's errors and setting them on a `User` record would have trouble due to there being different attribute names.

Schparky: Seems to work. Also, I just took a glance at the Yii code to confirm it doesn't check the value of the array key.

api.raml
Schparky: Changed from 422 to 409.

application/common/models/Password.php
```diff
@@ -194,17 +197,13 @@ public function getUser()
...
-     * @return bool
+     * @param $attribute
      * @throws ConflictHttpException
      */
-    public function save($runValidation = true, $attributeNames = null)
+    public function checkRecentlyUsed($attribute)
     {
         if ($this->hasAlreadyBeenUsedTooRecently()) {
-            throw new ConflictHttpException('May not be reused yet.', 1542395933);
+            throw new ConflictHttpException('May not be reused yet', 1542395933);
```
fillup:
```suggestion
            throw new ConflictHttpException('You have already used this password, please use a new password', 1542395933);
```

Schparky: This message is swallowed by `idp-pw-api`, and passes on the 409 to the UI with its own translated message.  In fact, as it stands today, the UI overrides the API message with its own.

PR #173, opened 2019-04-05
update welcome email subject and content

Schparky:


application/common/mail/welcome.html.php
```diff
@@ -28,68 +28,23 @@
...
-        Printable Codes)</li>
-    <li>Log out and log in again to see how it works and to have it remember your computer for 30 days. Note that
-        logging out will undo the "Remember this computer" setting.</li>
-</ol>
-<p>
-    To learn more about 2-Step Verification go to <?=yHtml::a(yHtml::encode($helpCenterUrl), $helpCenterUrl)?>
-</p>
-<?php endif; ?>
+    <li><strong>Username:</strong><?=yHtml::encode($username)?></li>
+    <li><strong>To update profile, go to:</strong><?=yHtml::encode($passwordProfileUrl)?></li>
```
fillup: Since this is an html email should we add anchor tags around the urls?

application/common/mail/welcome.text.php
```diff
@@ -25,57 +25,21 @@
 ?>
 Dear <?= $displayName ?>,
 
-Congratulations! You have logged into your new <?=$idpDisplayName . PHP_EOL ?>
-account for the first time. Please remember that this account will be your
-primary means for logging into many corporate applications. It is also
-important to note that this account (username and password) is unique and
-will not be kept in sync with any other accounts you have.
+<p>
```
fillup: Since this is the plain text email we need to remove all the html tags and just use whitespace for formatting.

PR #177, opened 2019-04-12
Add mfa and method nags

Schparky:
MFA nag works much like it did before being removed. It takes priority over recovery method nag and profile review. Recovery method nag takes priority over profile review, but not MFA nag. Dates for existing database records are set using the Yii `afterFind` event because it's impossible (or just very difficult) to do this properly in a migration.


application/common/models/User.php
```diff
@@ -521,22 +541,81 @@ public function getDisplayName()
...
+    /**
+     * Based on provided time, determine whether to present a reminder to add
+     * an MFA option.
+     * @param int $now
+     * @return bool
+     * @throws Exception
+     */
+    public function isTimeToNagToAddMfa(int $now): bool
+    {
+        return MySqlDateTime::isBefore($this->nag_for_mfa_after, $now)
```
wcjr: I'm finding `isBefore` confusing here...I would think the code would read "is now after the nag date"...is this code accomplishing that same thing but in some dvorak keyboard way?

application/features/authentication.feature
```diff
@@ -178,36 +178,32 @@ Feature: Authentication
...
-        | username  | shep_clark  |
-        | password  | govols!!!   |
-      And I request "/authentication" be created
-    When I request "/user/123" be retrieved
-    Then the response status code should be 200
-      And the following data is returned:
-        | property       | value      |
-        | username       | shep_clark |
-        | profile_review | no         |
+      | mfaTense | methodTense | reviewTense | mfaAdd | methodAdd | review |
```
wcjr: it's great to see the truth table expanded in this way, you've got a very complex situation here now.  This does not cover all possibilities though, is that because you were able to eliminate a bunch for logical reasons?

application/common/models/User.php
```diff
@@ -24,12 +24,20 @@ class User extends UserBase
     const SCENARIO_AUTHENTICATE    = 'authenticate';
     const SCENARIO_INVITE          = 'invite';
```
wcjr: would encapsulating these states in a class make the code neater?  It might also give you the opportunity to extract some of the state logic, e.g., determining current state, transitioning to another state, etc., into a more cohesive implementation.

application/common/models/User.php
```diff
@@ -521,22 +541,81 @@ public function getDisplayName()
...
-        return MySqlDateTime::isBefore($this->review_profile_after, time());
+        return MySqlDateTime::isBefore($this->review_profile_after, $now);
     }
 
+    /**
+     * Based on current time and presence of MFA and Method options,
+     * determine which "nag" to present to the user.
+     *
+     */
+    public function getNagState()
```
wcjr: see my notes earlier about the possibility of a NagState struct or class or something like that to encapsulate this kind of functionality.

application/common/models/User.php
```diff
@@ -521,22 +541,81 @@ public function getDisplayName()
...
+    }
+
+
     /**
      * @return array MFA related properties
      */
     public function getMfaFields()
     {
         return [
             'prompt'  => $this->isPromptForMfa() ? 'yes' : 'no',
```
wcjr: are `prompt` and `add` mutually exclusive?

application/console/migrations/m190411_201535_add_nag_dates_to_user_table.php
```diff
@@ -0,0 +1,27 @@
...
+ * Class m190411_201535_add_nag_dates_to_user_table
+ */
+class m190411_201535_add_nag_dates_to_user_table extends Migration
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function safeUp()
+    {
+        $this->addColumn('{{user}}', 'nag_for_mfa_after', 'date not null');
```
wcjr: I like that you included `after` here, very helpful!

application/common/models/User.php
Schparky: Yes, but I'm not sure what could be different. Qwerty is much more confusing! Seriously, the only thing I can think of is to make a new `isAfter` method and swap the two arguments.

application/features/authentication.feature
Schparky: What combination did I miss? There are eight combinations of three binary variables (2^3 = 8).

application/common/models/User.php
Schparky: I'll take a look. Thanks for the tip.

application/common/models/User.php
Schparky: No. That's what I though, too, at first. `prompt` tells `simplesamlphp-module-mfa` to present the user with a prompt for their *existing* mfa, whereas `add` tells `simplesamlphp-module-profilereview` to present the user with a reminder to add an `mfa` because they don't have one. Though they seem opposite, there are plenty of occasions when neither will be `yes`, simply because the `nag_for_mfa_after` date hasn't passed.

application/console/migrations/m190411_201535_add_nag_dates_to_user_table.php
Schparky: Thanks.

application/features/authentication.feature
wcjr: `yes | yes | yes` for example?

application/common/models/User.php
wcjr: a `MySqlDateTime.isAfter()` could be created and simply negate the other under the covers.  `MySqlDateTime::isAfter($now, $this->nag_for_mfa_after)` reads better to me but it's a nit if you don't think it's worth it.

application/common/models/User.php
wcjr: 👍

application/common/models/User.php
wcjr: 👍

application/features/authentication.feature
Schparky: Since `mfaAdd`, `methodAdd`, and `review` are outputs, that's not a valid test case.

wcjr: ok

PR #180, opened 2019-04-16
Remove LDAP password migration

Schparky:


wcjr:
I'm going to leave this review to @forevermatt or @fillup since they were closely involved with this functionality.

PR #191, opened 2019-05-09
Delay invite emails

Schparky:
### Changed
- Delay invite emails by 1 hour to allow time for a new email account to be created.
### Removed
- Removed `graced_period_ends_on` from API

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
-## [4.2.0] - 2019-05-03
+## [4.2.0] - 2019-05-06
 ### Added
 - Added recovery method reminder email, replacing daily method verify emails
 - Send an email notice when an unverified recovery email is purged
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [4.2.1] - 2019-05-09
+### Changed
+- Delay invite emails by 1 hour to allow time for a new email account to be created.
+### Removed
+- Removed `graced_period_ends_on` from API
```
forevermatt: Minor type: `graced` should be `grace`.

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [4.2.1] - 2019-05-09
+### Changed
+- Delay invite emails by 1 hour to allow time for a new email account to be created.
+### Removed
```
forevermatt: Would the removal of a field from the API constitute a backwards breaking change? I'm not trying to push for one way or the other, I just tend think of a "Removed" entry in the changelog resulting in a major version bump.

Schparky: I think technically you are right, and on some days I would be a stickler on things like this. Today, I chose to slip it in a patch because it's never used. If it's a bad idea, I'd rather just leave the API as it is rather than create a major version for a minor change.

PR #192, opened 2019-05-15
add multi-field search on /user endpoint

Schparky:


Schparky:
This change affects the inclusion of recovery method email addresses in all responses to `user` endpoints. Method data will be included in all cases, but email addresses will be masked out if not explicitly requested, or in an authentication scenario for the purpose of profile review. Prior behavior was to omit the data entirely except in the case of a profile review.

application/common/models/User.php
```diff
@@ -411,7 +413,14 @@ public function fields(): array
                 return Utils::getIso8601($model->last_login_utc);
             },
             'manager_email',
-            'personal_email',
+            'personal_email' => function (self $model) {
+                $maskParam = Yii::$app->request->queryParams['mask'] ?? 'yes';
```
fillup: Is there a use case where `Yii::$app->request` is not present? like in a cron job context?

Schparky: Good question. Yes, it is not present in console context, but the null coalesce operator handles that. I wasn't sure whether it just protected against a missing array key, or if it also protects against a missing object property, so I tested it and it works fine in the console.

fillup: Nice!

PR #194, opened 2019-05-28
Email template edits

Schparky:
- various edits requested during text review
- use `supportName` and `supportEmail` when `helpCenterUrl` is not provided
- provide for multi-line email signature


application/common/mail/method-verify.text.php
```diff
@@ -15,19 +15,23 @@
...
+following address to verify this email address:
 
 <?= $passwordProfileUrl . '/password/recovery/' . $uid . '/verify/' . $code . PHP_EOL ?>
 
-If you did not do this, it could be a sign someone else has compromised your
-account. Please contact <?= $supportName ?> at <?= $supportEmail ?> as soon
-as possible to report the incident.
+If you did not request adding this email address to your account please contact
+<?= $supportName ?> at <?= $supportEmail ?> as soon as possible to report the
+incident. Do NOT click the link above.
```
forevermatt: The HTML version of this email doesn't include the "Do NOT click the link above." Should it?

Schparky: Yes, thanks for catching that. Makes me want to remove the text templates.

PR #196, opened 2019-06-04
delete inactive users by cron task

Schparky:


application/console/controllers/CronController.php
```diff
@@ -129,4 +129,18 @@ public function actionSendPasswordExpiryEmails()
         $emailer->sendPasswordExpiringEmails();
         $emailer->sendPasswordExpiredEmails();
     }
+
+    public function actionDeleteInactiveUsers()
+    {
+        $logMessage = [ 'action' => 'delete inactive users' ];
+
+        if (\Yii::$app->params['inactiveUserDeletionEnable']) {
```
forevermatt: Would this conditional logic be better placed within the `deleteInactiveUsers()` method? That would prevent us from deleting inactive users whenever the config is set to false (which is more what people may expect when setting that configuration value), rather than just preventing the cron job from doing it.

That would also gather the logging all into the same place, rather than logging the start here in the cron controller and logging the end over in the User class.

Schparky: Good. This also allows for cleaner logic (i.e. removing an `if`).

Schparky: Good call. Also allows for cleaner logic (i.e. removing an `if`).

application/common/models/User.php
```diff
@@ -1026,4 +1030,56 @@ public function extendGracePeriodIfNeeded()
             ]);
         }
     }
+
+    /**
+     * Delete all user records that are inactive and have not been updated recently.
+     */
+    public static function deleteInactiveUsers()
+    {
+        $enabled = \Yii::$app->params['inactiveUserDeletionEnable'];
```
fillup: I think you forgot to test this value before proceeding, unless it is somewhere else?

Schparky: It's in the query, line 1060 (`andWhere('1=' . (int) $enabled)`)

forevermatt: That seems a little too subtle. What would you think of simply dealing with it right at the beginning?
```php
    if ($enabled !== true) {
        \Yii::warning([
            'action' => 'delete inactive users',
            'status' => 'disabled',
        ]);
        return;
    }
```

PR #212, opened 2019-07-31
cron deletes old manager mfa records

Schparky:
I originally thought `Mfa::removeOldUnverifiedRecords()` would take care of this, but it did not because manager codes are marked as verified when created.

### DIFF
--- a/application/common/models/Mfa.php
+++ b/application/common/models/Mfa.php
@@ -462,7 +462,7 @@ class Mfa extends MfaBase
      */
     protected static function deleteOldRecords(string $age, array $criteria): void
     {
-        $age = '-' . ltrim($age, '+');
+        $age = str_replace('+', '-', $age);
 
         /**
          * @var string $removeOlderThan  Records created before this date should be deleted
--- a/application/common/models/Mfa.php
+++ b/application/common/models/Mfa.php
@@ -462,6 +462,12 @@ class Mfa extends MfaBase
      */
     protected static function deleteOldRecords(string $age, array $criteria): void
     {
+        \Yii::warning([
+            'action' => 'delete old mfa records',
+            'criteria' => $criteria,
+            'status' => 'starting',
+        ]);
+
         $age = str_replace('+', '-', $age);
 
         /**
--- a/application/common/models/Mfa.php
+++ b/application/common/models/Mfa.php
@@ -453,19 +453,25 @@ class Mfa extends MfaBase
      */
     public static function removeOldUnverifiedRecords()
     {
-        /*
-         * Replace '+' with '-' so all env parameters can be defined consistently as '+n unit'
-         */
-        $mfaLifetime = str_replace('+', '-', \Yii::$app->params['mfaLifetime']);
+        self::deleteOldRecords(\Yii::$app->params['mfaLifetime'], ['verified' => 0]);
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


application/common/models/Mfa.php
```diff
@@ -453,19 +453,25 @@ public function getReadableType()
...
+        self::deleteOldRecords(\Yii::$app->params['mfaLifetime'], ['verified' => 0]);
+    }
+
+    /**
+     * @param string $age PHP relative time denoting how old a record must be to qualify for removal
+     * @param array $criteria array suitable for passing to a query, like ['field' => value]
+     */
+    protected static function deleteOldRecords(string $age, array $criteria): void
+    {
+        $age = '-' . ltrim($age, '+');
```
forevermatt: I think you'll want to keep this `+` to `-` replacement wherever you retrieve the `mfaLifetime` parameter (e.g. up in the `removeOldUnverifiedRecords()` method), so that this `deletedOldRecords()` method's `$age` argument will always be specified as `-something`.

I suspect that it'll also work out better to continue using the `str_replace('+', '-', ...)` approach rather than the `'-' . ltrim(..., '+')` approach, as the former can handle values specified with either a + or a -, but the latter can only handle values specified with a + (if there is a -, it would be doubled).

application/common/models/Mfa.php
```diff
@@ -566,4 +574,9 @@ public function setLabel($label)
             $this->label = $this->getReadableType();
         }
     }
+
+    public static function removeOldManagerMfaRecords(): void
+    {
+        self::deleteOldRecords('1 week', ['type' => Mfa::TYPE_MANAGER]);
```
forevermatt: I'm not sure I understand... this looks like it will not let anyone use the "manager" MFA option if their IdP account is more than a week old. Is that intentional? I could very well be misreading this, so please feel free to correct my understand of the code.

application/console/controllers/CronController.php
```diff
@@ -31,6 +31,12 @@ public function actionRemoveOldUnverifiedRecords()
             'status' => 'starting',
         ]);
         Mfa::removeOldUnverifiedRecords();
+
+        \Yii::warning([
```
forevermatt: Putting the "starting" and "complete" log message calls in separate functions makes it possible to call the code in such a way that only one of those messages will actually be logged. (Example: If any code outside of this CronController calls the `Mfa::removeOldManagerMfaRecords()` method, the "starting" log message wouldn't be sent.)

This isn't critical, but you may want to consider putting the "starting" and "complete" log messages in the same function, around the implementation (or sub call) that actually does what they're describing. Just a suggestion.

application/common/models/Mfa.php
Schparky: The date it's using is the `mfa` `created_utc` date, not the `user` created date (which, strangely, does not exist).

forevermatt: Okay. When is the "manager" MFA option added? I thought that was a built-in option added automatically as long as the user didn't have any other MFA options.

Schparky: It is added when the user selects "I need help" on the 2sv "More options" popup.

forevermatt: Ohhh! Thanks. This makes so much more sense now.

PR #213, opened 2019-08-07
improve Mfa::deleteOldRecords

Schparky:
The last edit before the previous release changed what types of relative dates Mfa::deleteOldRecords could handle. Specifically, it didn't correctly deal with '1 week', which caused all manager MFA records to be deleted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [4.6.3] - 2019-08-06
+### Added
```
forevermatt:
```suggestion
### Fixed
```

PR #221, opened 2020-01-21
deactivate a user that has had no email address for too long

Schparky:
Update a user record in the database to set `active` to `no` the first time the record is accessed after the expiration date. Prior to this change, the in-memory value of `active` was set to `no` but no change was made to the database. This resulted in correct API responses, but an internal inconsistency that created potential for confusion.

### DIFF
--- a/application/common/models/User.php
+++ b/application/common/models/User.php
@@ -878,6 +878,10 @@ class User extends UserBase
             $this->expires_on = null;
         }
 
+        if ($this->isExpired()) {
+            $this->deactivate();
+        }
+
         return parent::beforeSave($insert);
     }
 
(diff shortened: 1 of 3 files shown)


application/common/models/User.php
```diff
@@ -1115,4 +1111,34 @@ public static function numberWithOneMfaNotBackupCodes()
         ")->queryOne();
         return $result['COUNT(*)'];
     }
+
+    protected function isExpired(): bool
+    {
+        return $this->expires_on != null && MySqlDateTime::isBefore($this->expires_on, time());
```
forevermatt: Did you intentionally change `!==` to `!=` in the `expires_on` check?

application/common/models/User.php
```diff
@@ -1115,4 +1111,34 @@ public static function numberWithOneMfaNotBackupCodes()
...
+    }
+
+    protected function deactivate(): void
```
forevermatt: It may be worth clarifying some details about this method:

* Based on (new) line 1127, it looks like this is only intended to be used for deactivating an _expired_ user. If that's truly the only intended usage, what would you think about changing the name from `deactivate()` to `deactivateExpiredUser()`?
* I would normally expect a method to do what it says or, if it can't, then throw an exception. Given that this one simply logs any errors and returns normally (which is understandable, given when it's called), it may be worth documenting that behavior in a DocBlock above the method.

application/common/models/User.php
Schparky: Nope. Changed back to `!==`.

application/common/models/User.php
Schparky: Good thoughts. Updated.

PR #223, opened 2020-03-09
New configs to improve 2sv adoption

Schparky:
- `mfaAllowDisable` makes `require_mfa` a one-way switch
- `mfaRequiredForNewUsers` sets `require_mfa` to "yes" for all new users

Also added `require_mfa` property to User API responses, e.g. for https://github.com/silinternational/idp-id-broker-search.

### DIFF
--- a/application/features/bootstrap/UnitTestsContext.php
+++ b/application/features/bootstrap/UnitTestsContext.php
@@ -41,6 +41,9 @@ class UnitTestsContext extends YiiContext
     /** @var Invite */
     protected $inviteCode;
 
+    /** @var string[] */
+    protected $originalParams = [];
+
     /**
      * @afterScenario @database
      */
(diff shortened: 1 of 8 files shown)


application/features/bootstrap/UnitTestsContext.php
```diff
@@ -60,22 +63,25 @@ public function purgeDatabase()
      * username, they will be deleted.
      *
      * @param string $username
+     * @param string[] $properties
      * @return User
      */
-    protected function createNewUserInDatabase($username)
+    protected function createNewUserInDatabase($username, $properties = [])
     {
         $existingUser = User::findByUsername($username);
```
forevermatt: It might be worth overriding the `$username` with `$properties['username']` (if set), to protect against conflicting username data in the `$username` and `$properties` parameters. Otherwise, this initial check for (and removal of) any conflicting user won't be checking the correct username.

Something like this would probably work:
`$username = $properties['username'] ?? $username`

Schparky: Interesting observation. Would that go against the description in the method comment? "Create a new user in the database with the given username..." Maybe it would be better to unset a username in the properties array instead. What do you think?

forevermatt: Hmm, good catch. It would probably be best, then, to either assert that the `$properties` array doesn't contain a 'username' or add a note to the description that a 'username' in the `$properties` will override the value in the `$username` parameter. Either way seems fine.

PR #225, opened 2020-04-06
log to stdout/stderr for migration to CloudWatch logs

Schparky:
- send Yii logs to stdout/stderr
- send Apache logs to stdout/stderr (Note, this includes access logs. If access logs in CloudWatch turns out to be too much noise, it could be sent to /dev/null rather than collecting in a file in the container.)


application/composer.json
```diff
@@ -15,7 +15,8 @@
     "silinternational/yii2-email-log-target": "^1.0",
     "theiconic/php-ga-measurement-protocol": "^2.0",
     "yiisoft/yii2-swiftmailer": "^2.0",
-    "ramsey/uuid": "^3.6"
+    "ramsey/uuid": "^3.6",
+    "codemix/yii2-streamlog": "^1.3"
```
fillup, line 19: Does logging to stdout/err require a new external library?

Schparky: Yes, I was surprised to find that Yii2 doesn't have it built-in.

forevermatt: Would our psr3-adapters repo's [PSR3ConsoleLogger](https://github.com/silinternational/psr3-adapters/blob/develop/src/Psr3ConsoleLogger.php) be sufficient?

Schparky: Hmm, I'll look into it. Thanks for the suggestion.

Schparky: @forevermatt, I may be mistaken, but `PSR3ConsoleLogger` doesn't apply here because it is not a log target, but a PSR3 logger that intercepts the messages and `echo`s them before they are sent to the target. Did you have something else in mind to make this work?

forevermatt: Oh, thanks. You're correct: It's merely a PSR-3-compliant logger that simply echos to standard out, not a Yii Log Target.

PR #226, opened 2020-04-08
run a task if defined in env var RUN_TASK, otherwise start apache

Schparky:


application/console/controllers/CronController.php
```diff
@@ -116,4 +116,43 @@ public function actionDeleteInactiveUsers()
     {
         User::deleteInactiveUsers();
     }
+
```
fillup, line 119:
```suggestion
    // run all cron tasks, catching and logging errors to allow remaining tasks to run after an exception
```

PR #227, opened 2020-04-16
Check HIBP during login to auto-expire pwned passwords

fillup:
During login, if enabled, when the user's password is due to be checked with HaveIBeenPwned pwned passwords API  (default every 7 days) it'll check HIBP and if pwned:

1. Mark password as pwned
2. Expire password to require user to change on next login
3. Set grace period (default 7 days) so user can change password on login and not have to do Forgot Password immediately
4. Send the user an email letting them know what happened

If not pwned it'll extend the `password.check_hibp_after` date by the configured amount, default 7 days. 

If `HIBP_TRACKING_ONLY` is true it will only track to GA when pwned passwords are detected but will not mark passwords as pwned or notify the user. This is to allow us to learn / assess the impact before we start affecting users.


fillup:
> Maybe I missed it but I don't see where you avoid calling the HIBP API if password->hibp_is_pwned is already "yes".

Good point. I added a check for that. As you say, if we know it is pwned no need to check again.

application/common/models/Password.php
```diff
@@ -236,4 +236,40 @@ public function scenarios(): array
...
+     *  - set hibp_is_pwned to yes
+     *  - set expiration to now
+     *  - set graceperiod based on config
+     */
+    public function markPwned(): void
+    {
+        $this->setScenario(self::SCENARIO_UPDATE_METADATA);
+        $this->hibp_is_pwned = 'yes';
+        $this->expires_on = MySqlDateTime::relativeTime('-5 minutes');
+        $this->grace_period_ends_on = MySqlDateTime::relativeTime(\Yii::$app->params['hibpGracePeriod']);
```
Schparky, line 267: I think this will not work as intended for a user with MFA options. In such a case, the value returned to the API consumer is extended from the database value.

application/common/models/Password.php
```diff
@@ -271,5 +273,36 @@ public function markPwned(): void
                 'errors' => $this->getFirstErrors(),
             ]);
         }
+
+        // track event
+        try {
+            $trackingId = \Yii::$app->params['googleAnalytics']['trackingId']; // 'UA-12345678-12'
+            if ($trackingId === null) {
+                \Yii::warning(['google-analytics' => "Aborting GA cron, since the config has no GA trackingId"]);
```
Schparky, line 281:
```suggestion
                \Yii::warning(['google-analytics' => "Aborting GA pwned since the config has no GA trackingId"]);
```

application/common/models/Password.php
```diff
@@ -271,5 +273,36 @@ public function markPwned(): void
...
+                return;
+            }
+
+            $clientId = \Yii::$app->params['googleAnalytics']['clientId']; // 'IDP_ID_BROKER_LOCALHOST'
+            if ($clientId === null) {
+                \Yii::warning(['google-analytics' => "Aborting GA cron, since the config has no GA clientId"]);
```
Schparky, line 287:
```suggestion
                \Yii::warning(['google-analytics' => "Aborting GA pwned since the config has no GA clientId"]);
```

application/composer.json
```diff
@@ -15,11 +15,11 @@
...
-    "ramsey/uuid": "^3.6"
+    "ramsey/uuid": "^3.6",
+    "guzzlehttp/guzzle": "^6.2"
   },
   "require-dev": {
     "behat/behat": "^3.3",
-    "guzzlehttp/guzzle": "^6.2",
     "webmozart/assert": "^1.2",
     "friendsofphp/php-cs-fixer": "^2.14"
   },
```
forevermatt, line 25: When I install the composer dependencies, it says that the lock file is out of sync with the composer.json file, so you may need to do another `make depsupdate`.

application/features/authentication.feature
```diff
@@ -6,6 +6,20 @@ Feature: Authentication
...
+        | last_name    | User                  |
+        | display_name | Pwned User            |
+        | username     | pwned_user            |
+        | email        | pwned_user@example.org|
+      And I request "/user" be created
+      And a record exists with an employee_id of "93939202111"
+      And the user has a password of "pass123"
+
+    Given the requester is authorized
+
```
forevermatt, line 22: Extra blank line
```suggestion
```

application/features/bootstrap/fakes/HibpUnitTestsContext.php
```diff
@@ -0,0 +1,229 @@
+<?php
+namespace Sil\SilIdBroker\Behat\Context;
```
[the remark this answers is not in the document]
fillup: good catch, I wasn't sure what all I needed initially so I copied them all, I'll delete it.

application/features/bootstrap/fakes/HibpUnitTestsContext.php
```diff
@@ -0,0 +1,229 @@
+<?php
```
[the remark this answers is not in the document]
fillup: thanks, deleted it.

application/common/models/Password.php
fillup: oh good point. So should I set `expires_on` to something like 20 years ago to be safe?

Schparky: I don't think that would work. It's the grace period that matters, I think, and unless you check for the presence of a valid MFA, you won't know how to set `grace_period_ends_on`. Thinking right now there are a few options, (1) check for MFA and set the grace period accordingly, or (2) revise how the grace period is implemented so the stored grace period is actually what is returned to the consumer, or (3) leave it as it is knowing that a user with MFA is relatively safe even with a pwned password.

application/common/models/User.php
```diff
@@ -292,6 +294,52 @@ private function validatePassword(): Closure
...
+    /**
+     * If user is due to have password checked with HIBP, check it
+     * If password is found to be pwned, process it
+     * Fail gracefully to allow user to login if HIBP is unavailable
+     */
+    public function checkAndProcessHIBP(): void
+    {
+        if ( ! \Yii::$app->params['hibpCheckOnLogin'] ||
+            empty($this->password) ||
+            time() < strtotime($this->currentPassword->check_hibp_after)) {
```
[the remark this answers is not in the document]
fillup: sure, updated and will push with a batch of updates shortly.

application/composer.json
fillup: okay, I did that, will include in a batch of updates shortly.

README.md
```diff
@@ -47,3 +47,9 @@ email.
 ## API Documentation
 The API is described by [api.raml](api.raml), and an auto-generated [api.html](api.html) created by
 `raml2html`. To regenerate the HTML file, run `make raml2html`.
+
+## Running tests interactively locally
```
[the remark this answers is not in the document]
fillup: Sure, I added a new `make testcli` that does the first three steps and drops you in a shell. will push shortly.

application/features/bootstrap/FeatureContext.php
```diff
@@ -742,4 +742,24 @@ public function iWaitUntilAfterTheUserExpirationDate($username)
...
     }
+
+    /**
+     * @Given /^The user's current password should not be marked as pwned$/
+     */
+    public function theUserSCurrentPasswordShouldNotBeMarkedAsPwned()
+    {
+        $user = User::findOne(['username' => $this->reqBody['username']]);
+        Assert::eq($user->currentPassword->hibp_is_pwned, "no");
+        Assert::greaterThanEq(strtotime($user->currentPassword->expires_on), time());
```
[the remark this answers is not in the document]
fillup: Okay, I moved these to separate functions for checking if expired or not expired.

application/features/bootstrap/FeatureContext.php
```diff
@@ -742,4 +742,24 @@ public function iWaitUntilAfterTheUserExpirationDate($username)
         $this->userFromDb->scenario = User::SCENARIO_UPDATE_USER;
         $this->userFromDb->save();
     }
+
+    /**
+     * @Given /^The user's current password should not be marked as pwned$/
```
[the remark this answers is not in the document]
fillup: I removed the `/^` and `$/` and it worked fine.

application/features/bootstrap/HibpUnitTestsContext.php
```diff
@@ -0,0 +1,78 @@
...
+        Assert::eq($this->hashes['Prefix'], $hashPrefix);
+        Assert::eq($this->hashes['Suffix'], $hashSuffix);
+    }
+
+    /**
+     * @Given I have a pwned :arg1
+     */
+    public function iHaveAPwned($arg1)
+    {
+        $this->password = $arg1;
```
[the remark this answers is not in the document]
fillup: sure, I changed it to be that way

application/common/models/Password.php
fillup: Thanks @Schparky that's good looking out! What about this solution I thought of?

```php
    /**
     * If password is pwned, return actual expires_on.
     * Otherwise calculate expires_on date based on if user has MFA configured
     * @return string
     */
    public function getExpiresOn()
    {
        return $this->hibp_is_pwned == 'yes' ? $this->expires_on : $this->getMfaExtendedDate($this->expires_on);
    }

    /**
     * If password is pwned, return actual grace_period_ends_on.
     * Otherwise calculate grace_period_ends_on based on if user has MFA configured
     * @return string
     */
    public function getGracePeriodEndsOn()
    {
        return $this->hibp_is_pwned == 'yes' ? $this->grace_period_ends_on : $this->getMfaExtendedDate($this->grace_period_ends_on);
    }
```

Since these functions already existed inside `Password` it seemed reasonable to take into account `$this->hibp_is_pwned`.

Schparky: Looks like a perfect solution!

application/common/models/User.php
```diff
@@ -319,33 +320,78 @@ public function checkAndProcessHIBP(): void
...
-            ]);
+        if (! $this->isPasswordPwned()) {
+            $this->currentPassword->extendHibpCheckAfter();
+            return;
+        }
 
+        if (\Yii::$app->params['hibpTrackingOnly']) {
+            // extend check after date to only track user once per checking period
+            $this->currentPassword->extendHibpCheckAfter();
+            $this->trackPwnedPasswordGAEvent();
```
Schparky, line 331: Can the `trackPwnedPasswordGAEvent` call be moved above the `if (\Yii::$app->params['hibpTrackingOnly'])` and line 336 be removed, since it's called regardless of the tracking-only param?

fillup: sure, will do.

application/common/mail/password-pwned.html.php
```diff
@@ -18,7 +18,7 @@
 <p>
     This email is to inform you that the password you use with your <?= $idpDisplayName ?> identity (IdP) account has
     been discovered in a database of leaked passwords. This does not specifically mean your <?= $idpDisplayName ?>
```
forevermatt, line 20: Is `$idpDisplayName` more like "Acme" or "Acme IdP"?

If the former, we may want to change `your <?= $idpDisplayName ?> account` to `your <?= $idpDisplayName ?> identity account`, to help protect against people misinterpreting this to be referring to their corporate email account's password.

application/common/models/User.php
```diff
@@ -294,48 +295,105 @@ private function validatePassword(): Closure
...
+     *  - if $this->password is not empty
+     *
+     */
+    public function shouldHibpBeChecked(): bool
+    {
+        return (\Yii::$app->params['hibpCheckOnLogin'] &&
+                ! empty($this->password) &&
+                time() >= strtotime($this->currentPassword->check_hibp_after) &&
+                $this->currentPassword->hibp_is_pwned == 'no');
+    }
```
forevermatt, line 310: Excellent improvement! This seems much easier to read, including the change from ||'s to &&'s.

application/common/models/User.php
```diff
@@ -294,48 +295,105 @@ private function validatePassword(): Closure
...
+            $emailer->sendMessageTo(EmailLog::MESSAGE_TYPE_PASSWORD_PWNED, $this, [
+                'bccAddress' => \Yii::$app->params['hibpNotificationBcc']
+            ]);
         } catch (EmailServiceClientException $e) {
+            \Yii::error([
+                    'action' => 'check and process hibp',
+                    'employee_id' => $this->employee_id,
+                    'message' => 'unable to send password-pwned email to user',
+                    'error' => $e->getMessage(),
+                ]);
```
forevermatt, line 351: Indentation fix:

```suggestion
            \Yii::error([
                'action' => 'check and process hibp',
                'employee_id' => $this->employee_id,
                'message' => 'unable to send password-pwned email to user',
                'error' => $e->getMessage(),
            ]);
```

application/common/mail/password-pwned.html.php
fillup: yeah, and to be consistent with the use right above it. I'll identity in that second instance.

PR #228, opened 2020-04-20
Logging to stdout/err and new feature to check HIBP during login

fillup:
Should this be a minor or major upgrade? I don't think there is any compatibility changes, just a new feature and a change to logging.


application/common/config/main.php
```diff
@@ -225,5 +221,10 @@
         'mfaManagerHelpBcc'             => Env::get('MFA_MANAGER_HELP_BCC', ''),
         'contingentUserDuration'        => Env::get('CONTINGENT_USER_DURATION', '+4 weeks'),
         'inviteEmailDelaySeconds'       => Env::get('INVITE_EMAIL_DELAY_SECONDS', 0),
+        'hibpCheckOnLogin'              => Env::get('HIBP_CHECK_ON_LOGIN', true),
+        'hibpCheckInterval'             => Env::get('HIBP_CHECK_INTERVAL', '+1 week'),
+        'hibpGracePeriod'               => Env::get('HIBP_GRACE_PERIOD', '+1 week'),
+        'hibpTrackingOnly'              => Env::get('HIBP_TRACKING_ONLY', false),
```
Baggerone, line 227: Do you have a process in mind to ensure that we set this to `true` when it goes live? (That's what you said Sheryl wants, right?)

application/common/mail/password-pwned.html.php
```diff
@@ -0,0 +1,53 @@
...
+
+<p>
+    This email is to inform you that the password you use with your <?= $idpDisplayName ?> identity (IdP) account has
+    been discovered in a database of leaked passwords. This does not specifically mean your <?= $idpDisplayName ?>
+    identity account has been hacked, it just means you (or someone else) used the same password on other websites as
+    well, and one of them was hacked and had its database of credentials stolen.
+</p>
+
+<p>
+    As a result, your <?= $idpDisplayName ?> account password has been disabled for future use and you will need to
```
forevermatt, line 26:
```suggestion
    As a result, your <?= $idpDisplayName ?> identity account password has been disabled for future use and you will need to
```

application/common/config/main.php
fillup: Yeah, we'll have to run through TF updates first and add the variable. Since it is a temporary thing that we may not use for all IdPs I defaulted to `false` the way we want it to really work.

application/common/models/User.php
```diff
@@ -292,6 +295,110 @@ private function validatePassword(): Closure
...
+        }
+
+        $this->currentPassword->markPwned();
+
+        // notify user
+        try {
+            /* @var $emailer Emailer */
+            $emailer = \Yii::$app->emailer;
+            $emailer->sendMessageTo(EmailLog::MESSAGE_TYPE_PASSWORD_PWNED, $this, [
+                'bccAddress' => \Yii::$app->params['hibpNotificationBcc']
```
forevermatt, line 343: When in "track-only" mode, were we supposed to also send the email to the `hibpNotificationBcc` recipient?

For some reason I thought so, but perhaps "track-only" was merely intended to gather GA analytics, not emails to the IT support staff about which accounts are using a pwned password. Or does the GA analytics call individually identify the account whose password was leaked?

fillup: tracking only just gets numbers. it is intended just to get a sense of scale of the problem before support is possibly overwhelmed.

application/console/controllers/CronController.php
```diff
@@ -116,4 +116,46 @@ public function actionDeleteInactiveUsers()
     {
         User::deleteInactiveUsers();
     }
+
+    /**
+     * Run all cron tasks, catching and logging errors to allow remaining tasks to run after an exception
+     */
+    public function actionAll()
```
forevermatt, line 123: Since we're changing what cron commands need to be in the skinny repos that use ID Broker, it seems like this might be a major version bump.

forevermatt: Oh, never mind. I see that this is merely adding a new way, not yet removing the old way these commands could be called.

It's probably worth documenting that the old way is deprecated (if we plan to remove it in some future version).

fillup: sure, I don't know if the actual "api" changes, just how we're invoking it, but it's not a problem to do a major version change

fillup:
> Just noticed the new variables are not in `local.env.dist`.
> 
> ```
> # === haveibeenpwned configs ===
> 
> # Check password against hibp on login
> HIBP_CHECK_ON_LOGIN=true
> 
> # Check interval
> HIBP_CHECK_INTERVAL=+1 week
> 
> # Grace period to allow user to change password
> HIBP_GRACE_PERIOD=+1 week
> 
> # Set to true to track hibp but not expire leaked passwords
> HIBP_TRACKING_ONLY=false
> 
> # Email address to bcc pwned alerts
> HIBP_NOTIFICATION_BCC=
> ```

Thanks, I added them.

PR #233, opened 2020-04-28
limit password expired emails

Schparky:
Since at least one IdP instance was affected by an out-of-memory problem on sending password expiry messages, we realize it is not helpful to send password expired messages long after the fact. This change limits these messages to only those that have a password expired in the past 15 days.

### DIFF
--- a/application/common/components/Emailer.php
+++ b/application/common/components/Emailer.php
@@ -718,12 +718,15 @@ class Emailer extends Component
         foreach ($users as $user) {
             /** @var Password $userPassword */
             $userPassword = $user->currentPassword;
-            if ($userPassword
-                && strtotime($userPassword->getExpiresOn()) < time()
-                && ! $this->hasReceivedMessageRecently($user->id, EmailLog::MESSAGE_TYPE_PASSWORD_EXPIRED)
-            ) {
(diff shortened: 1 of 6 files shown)


application/common/components/Emailer.php
```diff
@@ -694,24 +703,36 @@ public function sendPasswordExpiredEmails()
...
-            if ($userPassword
-                && strtotime($userPassword->getExpiresOn()) < time()
-                && ! $this->hasReceivedMessageRecently($user->id, EmailLog::MESSAGE_TYPE_PASSWORD_EXPIRED)
-            ) {
-                $this->sendMessageTo(EmailLog::MESSAGE_TYPE_PASSWORD_EXPIRED, $user);
-                $numEmailsSent++;
+            if ($userPassword) {
+                $passwordExpiry = strtotime($userPassword->getExpiresOn());
+                if ($passwordExpiry < time()
+                    && $passwordExpiry > strtotime('-15 days')
```
Baggerone, line 724: Not a big deal and maybe not worth it at this stage.  I was thinking these strtotime values (here and line 681) should probably be constants and maybe even configurable with a default.  (If we make them configurable, though, we should probably include a validator that ensures there's no typo or something that could sabotage it.)

Schparky: I'll change to constants, but don't want to mess with config vars at this point.

PR #234, opened 2020-04-29
Release 5.0.2 -- limit password-expired notifications

Schparky:


application/common/components/Emailer.php
```diff
@@ -694,24 +706,36 @@ public function sendPasswordExpiredEmails()
...
-            if ($userPassword
-                && strtotime($userPassword->getExpiresOn()) < time()
-                && ! $this->hasReceivedMessageRecently($user->id, EmailLog::MESSAGE_TYPE_PASSWORD_EXPIRED)
-            ) {
-                $this->sendMessageTo(EmailLog::MESSAGE_TYPE_PASSWORD_EXPIRED, $user);
-                $numEmailsSent++;
+            if ($userPassword) {
+                $passwordExpiry = strtotime($userPassword->getExpiresOn());
+                if ($passwordExpiry < time()
+                    && $passwordExpiry > strtotime(self::FIFTEEN_DAYS_AGO)
```
Baggerone, line 727: You don't need to change it at this point, but just to clarify, I was thinking more along the lines of a constant that describes what this is for, e.g. PASSWORD_EXPIRY_OUTDATED_CUTOFF (not putting too much thought into it)

application/common/components/Emailer.php
```diff
@@ -54,6 +54,9 @@ class Emailer extends Component
     const PROP_TEXT_BODY = 'text_body';
     const PROP_DELAY_SECONDS = 'delay_seconds';
 
+    const PASSWORD_EXPIRED_CUTOFF = '-15 days';
```
Baggerone, line 57: :+1:

PR #236, opened 2020-05-26
Google Sheets export

Schparky:


application/common/models/User.php
```diff
@@ -1274,4 +1276,33 @@ public static function getActiveUnlockedUsers()
...
+
+        $googleSheetsClient = new Sheets([
+            'applicationName' => \Yii::$app->params['google']['applicationName'],
+            'jsonAuthFilePath' => \Yii::$app->params['google']['jsonAuthFilePath'],
+            'jsonAuthString' => \Yii::$app->params['google']['jsonAuthString'],
+            'delegatedAdmin' => \Yii::$app->params['google']['delegatedAdmin'],
+            'spreadsheetId' => \Yii::$app->params['google']['spreadsheetId'],
+        ]);
+
+        $activeUsers = User::find()->where(['active' => 'yes', 'locked' => 'no'])->all();
```
fillup, line 1295:
```suggestion
        $activeUsers = User::find()->where(['active' => 'yes'])->all();
```

fillup: I think we can/should include locked users since that can just be a temporary thing.

PR #245, opened 2021-02-15
Add ability to email regarding abandoned users

jason-jackson:
Add capabilities to email a contact person regarding inactive users.

### DIFF
--- a/application/common/components/Emailer.php
+++ b/application/common/components/Emailer.php
@@ -46,7 +46,7 @@ class Emailer extends Component
     const SUBJ_PASSWORD_EXPIRED = 'The password for your {idpDisplayName} Identity account has expired';
     const SUBJ_PASSWORD_PWNED = 'ALERT: The password for your {idpDisplayName} Identity account has been exposed';
 
-    const SUBJ_UNUSED_USER_ACCOUNTS = 'Unused SIL Google and SIL Identity Accounts';
+    const SUBJ_UNUSED_USER_ACCOUNTS = 'Unused SIL Identity Accounts';
 
     const PROP_SUBJECT = 'subject';
     const PROP_TO_ADDRESS = 'to_address';
--- a/application/common/components/Emailer.php
+++ b/application/common/components/Emailer.php
@@ -138,7 +138,10 @@ class Emailer extends Component
     public $subjectForPasswordExpired;
     public $subjectForPasswordPwned;
 
-    public $subjectForAbandonedUserAccounts;
+    public $subjectForAbandonedUsers;
+
+    /* The email to contact for HR notifications */
+    public $hrNotificationsEmail;
 
     /* The number of days of not using a security key after which we email the user */
     public $lostSecurityKeyEmailDays;
@@ -307,7 +310,7 @@ class Emailer extends Component
         $this->subjectForPasswordExpired = $this->subjectForPasswordExpired ?? self::SUBJ_PASSWORD_EXPIRED;
         $this->subjectForPasswordPwned = $this->subjectForPasswordPwned ?? self::SUBJ_PASSWORD_PWNED;
 
-        $this->subjectForAbandonedUserAccounts = $this->subjectForAbandonedUserAccounts ?? self::SUBJ_ABANDONED_USER_ACCOUNTS;
+        $this->subjectForAbandonedUsers = $this->subjectForAbandonedUsers ?? self::SUBJ_ABANDONED_USER_ACCOUNTS;
 
         $this->subjects = [
             EmailLog::MESSAGE_TYPE_INVITE => $this->subjectForInvite,
@@ -330,7 +333,9 @@ class Emailer extends Component
             EmailLog::MESSAGE_TYPE_PASSWORD_EXPIRED => $this->subjectForPasswordExpired,
             EmailLog::MESSAGE_TYPE_PASSWORD_PWNED => $this->subjectForPasswordPwned,
         ];
-        
+
+        $hrNotificationsEmail = $this->hrNotificationsEmail ?? '';
+
         $this->assertConfigIsValid();
         
         $this->verifyOtherDataForEmailIsValid();
@@ -761,7 +766,7 @@ class Emailer extends Component
     
             $htmlBody = \Yii::$app->view->render('@common/mail/abandoned-users.html.php', $dataForEmail);
     
-            $this->email($dataForEmail['hrNotificationsEmail'], $this->subjectForAbandonedUserAccounts, $htmlBody, strip_tags($htmlBody));
+            $this->email($dataForEmail['hrNotificationsEmail'], $this->subjectForAbandonedUsers, $htmlBody, strip_tags($htmlBody));
         }
     }
 }
--- a/application/common/components/Emailer.php
+++ b/application/common/components/Emailer.php
@@ -763,11 +763,10 @@ class Emailer extends Component
 
         if (!empty($this->hrNotificationsEmail)) {
             $dataForEmail['users'] = User::getAbandonedUsers();
-            
+
             if (!empty($dataForEmail['users'])) {
-                echo "Found ".count($dataForEmail['users'])." Users".PHP_EOL;
                 $htmlBody = \Yii::$app->view->render('@common/mail/abandoned-users.html.php', $dataForEmail);
-    
+
                 $this->email($this->hrNotificationsEmail, $this->subjectForAbandonedUsers, $htmlBody, strip_tags($htmlBody));
             }
         }
--- a/application/common/components/Emailer.php
+++ b/application/common/components/Emailer.php
@@ -334,7 +334,7 @@ class Emailer extends Component
             EmailLog::MESSAGE_TYPE_PASSWORD_PWNED => $this->subjectForPasswordPwned,
         ];
 
-        $hrNotificationsEmail = $this->hrNotificationsEmail ?? '';
+        $this->hrNotificationsEmail = $this->hrNotificationsEmail ?? '';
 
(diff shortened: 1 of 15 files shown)


application/common/config/main.php
```diff
@@ -205,8 +205,14 @@
         'passwordGracePeriodExtension'  => '+7 days',
         'inviteLifespan'                => Env::get('INVITE_LIFESPAN', '+1 month'),
         'inviteGracePeriod'             => Env::get('INVITE_GRACE_PERIOD', '+3 months'),
-        'inactiveUserDeletionEnable'    => Env::get('INACTIVE_USER_DELETION_ENABLE', false),
-        'inactiveUserPeriod'            => Env::get('INACTIVE_USER_PERIOD', '+18 months'),
+        'inactiveUser'                  => [
```
Schparky, line 208: You could use `Env::getArrayFromPrefix` if you either rename `DEACTIVATE_INSTRUCTIONS_URL` or move it outside the `inactiveUser` block. (Of course, the latter would require a change to `Emailer` as well.)  There are examples of this near the top of `main.php`. One benefit is the code saved here, but there are other potential benefits in the Terraform code if all related variables have a common prefix.

application/common/mail/unused-users.html.php
```diff
@@ -0,0 +1,57 @@
...
+ * @var string $deactivateInstructionsUrl
+ * @var array  $users
+ * @var string $emailSignature
+ */
+?>
+<p>Dear <?= yHtml::encode($contactName) ?>,</p>
+
+<p>
+    As GTIS works towards securing SIL's accounts, we are auditing SIL Identity access and SIL Google accounts and
+    asking HR to consider deactivating accounts that haven't been used in more than 6 months.
```
Schparky, line 16: Should this text ("6 months") match `INACTIVE_USER_PERIOD`?

application/common/components/Emailer.php
```diff
@@ -323,6 +329,7 @@ public function init()
             EmailLog::MESSAGE_TYPE_PASSWORD_EXPIRING => $this->subjectForPasswordExpiring,
             EmailLog::MESSAGE_TYPE_PASSWORD_EXPIRED => $this->subjectForPasswordExpired,
             EmailLog::MESSAGE_TYPE_PASSWORD_PWNED => $this->subjectForPasswordPwned,
+            EmailLog::MESSAGE_TYPE_UNUSED_USERS => $this->subjectForUnusedUserAccounts,
```
Schparky, line 332: I suggest defining this constant (`MESSAGE_TYPE_UNUSED_USERS`) here in `Emailer`. See comment in `EmailLog.php` for explanation.

application/common/models/EmailLog.php
```diff
@@ -31,6 +31,7 @@ class EmailLog extends EmailLogBase
     const MESSAGE_TYPE_PASSWORD_EXPIRING = 'password-expiring';
     const MESSAGE_TYPE_PASSWORD_EXPIRED = 'password-expired';
     const MESSAGE_TYPE_PASSWORD_PWNED = "password-pwned";
+    const MESSAGE_TYPE_UNUSED_USERS = 'unused-users';
```
Schparky, line 34: I suggest to not to define this here since it is not a logged email type. Maybe even better would be to proceed with this implementation by adding this type to the database enum and find a way to actually log this activity. (Log with a null user ID for example.)

application/common/models/User.php
```diff
@@ -1183,6 +1183,29 @@ public static function deleteInactiveUsers()
         ]);
     }
 
+
+    /**
+     * Delete all user records that are inactive and have not been updated recently.
```
Schparky, line 1188:
```suggestion
     * Get all user records that are inactive and have not been updated recently.
```

application/common/models/User.php
```diff
@@ -1183,6 +1183,29 @@ public static function deleteInactiveUsers()
...
+     */
+    public static function getInactiveUsers()
```
Schparky, line 1190: Looks like you could (should) use `getInactiveUsers` in `deleteInactiveUsers` assuming the difference is insignificant. It's confusing to me without greater analysis why the two queries are different (`last_changed_utc` vs `last_login_utc` + `created_utc`). If they ARE different, it would be less confusing to rename this function (and possibly the other as well) to describe why they are different.

application/console/controllers/CronController.php
```diff
@@ -117,6 +117,13 @@ public function actionDeleteInactiveUsers()
         User::deleteInactiveUsers();
     }
 
+    public function actionSendInactiveUsersEmail()
+    {
+        /* @var $emailer Emailer */
+        $emailer = \Yii::$app->emailer;
+        $emailer->actionSendInactiveUsersEmail();
```
Schparky, line 124:
```suggestion
        $emailer->sendInactiveUsersEmail();
```

application/console/controllers/CronController.php
```diff
@@ -117,6 +117,13 @@ public function actionDeleteInactiveUsers()
         User::deleteInactiveUsers();
     }
 
+    public function actionSendInactiveUsersEmail()
+    {
+        /* @var $emailer Emailer */
```
Schparky, line 122: Unless I missed it, we probably need a way to bail out if `INACTIVE_USER_CONTACT_EMAIL` isn't defined. I'm assuming that at least one of the IDPs will not want this feature.

application/console/controllers/CronController.php
```diff
@@ -117,6 +117,13 @@ public function actionDeleteInactiveUsers()
         User::deleteInactiveUsers();
     }
 
+    public function actionSendInactiveUsersEmail()
```
Schparky, line 120: Add `actionSendInactiveUsersEmail` to `actionAll`. We recently change the cron implementation and no longer have the flexibility to choose different schedules for each cron action.

application/common/components/Emailer.php
```diff
@@ -46,6 +46,8 @@ class Emailer extends Component
     const SUBJ_PASSWORD_EXPIRED = 'The password for your {idpDisplayName} Identity account has expired';
     const SUBJ_PASSWORD_PWNED = 'ALERT: The password for your {idpDisplayName} Identity account has been exposed';
 
+    const SUBJ_UNUSED_USER_ACCOUNTS = 'Unused SIL Identity Accounts';
```
Schparky, line 49:
```suggestion
    const SUBJ_UNUSED_USER_ACCOUNTS = 'Unused {idpDisplayName} Identity Accounts';
```

Schparky (changes requested):
Good work. There are a few things to address, but I like what you've done.

Baggerone (approved):

Baggerone (approved):

application/common/config/main.php
Schparky: After our Slack conversation, I decided it's not worth making this change. As you pointed out, two existing variables would break backward compatibility. Let's keep it as it is.

application/console/controllers/CronController.php
jason-jackson: Should `INACTIVE_USER_CONTACT_EMAIL` be used? I think there was some discussion to use `hr_notifications_email` or `notifier_email_to` instead?

Schparky: I think so. Seems the consensus was `HR_NOTIFICATIONS_EMAIL`. For the sake of clarity, it should be all-caps when specified as an environment variable. In Terraform, it is all lowercase. I was in the midst of Terraform work when the other conversation happened so that's why I used lowercase.

Let's keep it simple and not add the `bool` unless it's later warranted.

application/common/mail/unused-users.html.php
jason-jackson: Fixed this as well as adding in `idpDisplayName` where needed.

application/common/models/EmailLog.php
jason-jackson: Moved out of EmailLog for now.

Schparky (approved):
All that remains of my suggestions is to consolidate duplicate code in `getInactiveUsers` but that's not a necessity. Also, it would be good to look at adding some automated tests to go with these changes.

fillup (approved):

application/common/models/User.php
```diff
@@ -1183,6 +1183,29 @@ public static function deleteInactiveUsers()
...
+
+        /**
+         * @var string $removeBefore   All records that have not been updated since before this date
+         * should be deleted. Calculated relative to now (time of execution).
+         */
+        $removeBefore = MySqlDateTime::relative($inactiveUserPeriod);
+        return self::find()
+            ->andWhere(['<', 'last_login_utc', $removeBefore])
+            ->andWhere(['<', 'created_utc', $removeBefore])
+            ->andWhere(['active' => 'no'])
```
jason-jackson, line 1205: @fillup Is this correct or should it be `yes` instead?

application/common/models/User.php
jason-jackson: After talking with Phillip, methods should be different, and I renamed them.

application/common/models/User.php
jason-jackson: Based on Slack conversation, I changed this.

Schparky (approved):

local.env.dist
```diff
@@ -123,6 +123,12 @@ GA_CLIENT_ID=
 #SUBJECT_FOR_METHOD_PURGED=
 #SUBJECT_FOR_PASSWORD_EXPIRING=
 #SUBJECT_FOR_PASSWORD_EXPIRED=
+#SUBJECT_FOR_PASSWORD_EXPIRED=
```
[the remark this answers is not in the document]
jason-jackson: Done

PR #247, opened 2021-03-03
Track (changes to) root-level dependencies

forevermatt:


Makefile
```diff
@@ -12,8 +12,12 @@ bash:
 deps:
 	docker-compose run --rm cli composer install
 
+depsshow:
```
Schparky, line 15: clever way to split this out

Schparky (approved):

PR #252, opened 2021-11-08
Allow hyphen in employee id for UUIDs

Baggerone:


jason-jackson (commented):
I would add tests for UUIDs,  not change old tests as I think we'd want both to work.

application/features/user.feature
```diff
@@ -433,7 +433,7 @@ Feature: User
       And the user store is empty
       And I provide the following valid data:
         | property     | value                 |
-        | employee_id  | 123-456               |
+        | employee_id  | 123&456               |
```
forevermatt, line 436: I think it would be better to leave out the `employee_id` on this line (since this step is "I provide the following valid data:") and instead add a new line after this table (around line 441) that's something like this:
```
But I provide an invalid employee_id of "123&456"
```

forevermatt (changes requested):
Minor change requested

Baggerone:
There is already another test that creates a user with a simple employee_id.

application/features/user.feature
Baggerone: :+1:

forevermatt (approved):

PR #256, opened 2021-11-10
Release 5.3.3 Avoid error spam by not checking the email service status

Baggerone:


application/frontend/controllers/SiteController.php
```diff
@@ -77,8 +86,6 @@ public function actionStatus()
             ]);
             throw new Http500('Email Service problem.', $e->getCode());
         }
-
```
mtompset, line 80: Why did you remove this block from the function? Is the function even called anymore?

application/frontend/controllers/SiteController.php
```diff
@@ -40,14 +40,23 @@ public function actionStatus()
...
         } catch (Exception $e) {
             Yii::error('Emailer config problem: ' . $e->getMessage());
             throw new Http500('Emailer config problem.');
         }
-        
+
+        // Checking the email service status causes too many error emails to be sent out to the dev team
+        // $this->checkEmailServiceStatus($emailer);
+
+        Yii::$app->response->statusCode = 204;
```
mtompset, line 55: Had to look up 204, but that seems okay.
https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

application/frontend/controllers/SiteController.php
Baggerone: I'm not sure what you're referring to.  The git diff makes things look funny, because it shows the `statusCode = 204` as having been added above and removed down here, where as to the human eye it's still in the same place it was before inside its original function.

application/frontend/controllers/SiteController.php
mtompset: Also, this is the better placement for this line of code, since all the other checks are try/catch. It's weird that a private function sets the return value for this function.

application/frontend/controllers/SiteController.php
mtompset: Nevermind. I figured it out. Exactly.

mtompset (approved):
Not massively familiar with this system, but the code change looks simple enough.

forevermatt (approved):

PR #265, opened 2022-07-13
Release 6.0.2 (again but with updated changelog) Ensure clientExtensionResults is encoded as an object, even if empty

Baggerone:


Schparky (approved):
Honestly, I'd be fine with removing the CHANGELOG file.

PR #266, opened 2022-07-19
Not for merging ... delete webauthn credential

Baggerone:
Am I headed in the right direction on this?

Do you have any idea why I'm getting this error in my tiny behat test? (I tried adding the webauthn env vars in docker_compose.yml, but that didn't help.)

  When I request to delete a credential of the MFA with a credential_id of "123" # Sil\SilIdBroker\Behat\Context\MfaContext::iRequestToDeleteACredentialOfTheMfaWithACredentialIDOf()
  
  'message' => 'Typed property common\\components\\MfaBackendWebAuthn::$apiBaseUrl must not be accessed before initialization'

### DIFF
--- a/application/common/components/MfaBackendInterface.php
+++ b/application/common/components/MfaBackendInterface.php
@@ -35,4 +35,16 @@ interface MfaBackendInterface
      * @return bool
      */
     public function delete(int $mfaId): bool;
+
+
+
+    /**
+     * Delete WebAuthn credential
+     * @param int $mfaId
+     * @param string $credId
+     * @param string $rpOrigin
+     * @return bool
+     */
+    public function deleteCredential(int $mfaId, string $credId, string $rpOrigin): bool;
+
 }
--- a/application/common/components/MfaBackendManager.php
+++ b/application/common/components/MfaBackendManager.php
@@ -110,4 +110,17 @@ class MfaBackendManager extends Component implements MfaBackendInterface
     {
         return MfaBackupcode::deleteCodesForMfaId($mfaId);
     }
+
+
+    /**
+     * Delete WebAuthn credential
+     * @param int $mfaId
+     * @param string $credId
+     * @param string $rpOrigin
+     * @return bool
+     */
+    public function deleteCredential(int $mfaId, string $credId, string $rpOrigin): bool
+    {
+        return true;
+    }
 }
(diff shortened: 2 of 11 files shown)


application/frontend/controllers/MfaController.php
```diff
@@ -214,6 +215,41 @@ public function actionDelete(int $id)
...
+     * @throws \Throwable
+     * @throws BadRequestHttpException
+     * @throws NotFoundHttpException
+     * @throws ServerErrorHttpException
+     */
+    public function actionDeleteCredential(int $mfaId, string $credId)
+    {
+        $mfa = $this->getRequestedMfa($mfaId);
+
+        if ($mfa->type != Mfa::TYPE_WEBAUTHN) {
```
jason-jackson, line 232: I might throw this error in `deleteCredential` instead of returning true. Then delete this if statement.

application/common/components/MfaBackendTotp.php
```diff
@@ -117,4 +117,17 @@ public function delete(int $mfaId): bool
 
         return true;
     }
+
+
+    /**
+     * Delete WebAuthn credential
```
jason-jackson, line 123:
```suggestion
     * Delete credential
```

application/common/components/MfaBackendManager.php
```diff
@@ -110,4 +110,17 @@ public function delete(int $mfaId): bool
     {
         return MfaBackupcode::deleteCodesForMfaId($mfaId);
     }
+
+
+    /**
+     * Delete WebAuthn credential
```
jason-jackson, line 116:
```suggestion
     * Delete credential
```

application/common/components/MfaBackendInterface.php
```diff
@@ -35,4 +35,16 @@ public function verify(int $mfaId, string $value, string $rpOrigin = '');
      * @return bool
      */
     public function delete(int $mfaId): bool;
+
+
+
+    /**
+     * Delete WebAuthn credential
```
jason-jackson, line 42: Not sure if this comment should be changed or not.

```suggestion
     * Delete credential
```

jason-jackson:
I'm also not sure whether to just overload the delete method (add credId and rpOrigin, like we did for verify and authInit), or if we want to add the new method.

Baggerone:
Was only for discussion

PR #267, opened 2022-08-01
Break out a mfa_webauthn child table to allow for individually deleting webauthn credentials

Baggerone:


application/common/components/MfaBackendManager.php
```diff
@@ -103,11 +104,18 @@ public function verify(int $mfaId, string $value, string $rpOrigin = ''): bool
...
+     * @param string $credId Credential ID (only used for WebAuthn)
      * @return bool
      * @throws ServerErrorHttpException
      */
-    public function delete(int $mfaId): bool
+    public function delete(int $mfaId, string $credId=''): bool
     {
+        if ($credId != '') {
+            throw new ForbiddenHttpException(
+                sprintf("May not delete a credential on a %s mfa type", Mfa::TYPE_MANAGER),
```
jason-jackson, line 115: Does `$this->type` work instead of `Mfa:...`?

application/common/components/MfaBackendBackupcode.php
```diff
@@ -73,11 +74,19 @@ public function verify(int $mfaId, string $value, string $rpOrigin = ''): bool
...
+     * @param string $credId Credential ID (only used for WebAuthn)
      * @return bool
      * @throws ServerErrorHttpException
      */
-    public function delete(int $mfaId): bool
+    public function delete(int $mfaId, string $credId=''): bool
     {
+        if ($credId != '') {
+            throw new ForbiddenHttpException(
+                sprintf("May not delete a credential on a %s mfa type", Mfa::TYPE_BACKUPCODE),
```
jason-jackson, line 85: Does `$this->type` work instead of `Mfa:...`

jason-jackson (approved):
Odd to have both go and php in here. Maybe we should port this over to go sometime ;)

application/frontend/config/main.php
```diff
@@ -60,11 +60,12 @@
...
-                'POST   mfa'                          => 'mfa/create',
-                'PUT    mfa/<id:\d+>'                 => 'mfa/update',
-                'POST   mfa/<id:\d+>/verify'          => 'mfa/verify',
-                'DELETE mfa/<id:\d+>'                 => 'mfa/delete',
+                'GET    user/<employeeId:\w+>/mfa'              => 'mfa/list',
+                'POST   mfa'                                    => 'mfa/create',
+                'PUT    mfa/<id:\d+>'                           => 'mfa/update',
+                'POST   mfa/<id:\d+>/verify'                    => 'mfa/verify',
+                'DELETE mfa/<id:\d+>'                           => 'mfa/delete',
+                'DELETE mfa/<mfaId:\d+>/credential/<credId:.+>' => 'mfa/delete-credential',
```
forevermatt, line 68: Would the `credId` also be numeric, like the `mfaId`? (e.g. `<credId:\d+>`)

codeship-services.yml
```diff
@@ -36,3 +37,38 @@ app:
         HR_NOTIFICATIONS_EMAIL: hr@example.com
         ABANDONED_USER_bestPracticeUrl: http://www.example.com/best-practices.html
         ABANDONED_USER_deactivateInstructionsUrl: http://www.example.com/deactivate-instructions.html
+        MFA_WEBAUTHN_apiBaseUrl: externalapi:8080/
+        MFA_WEBAUTHN_apiKey: EC7C2E16-5028-432F-8AF2-A79A64CF3BC1
```
forevermatt, line 41: Are these dummy values or real values? They look real, so it seemed worth asking. If they are dummy values, would it be easy to change some segment of them to make that obvious? (e.g. changing `EC7C2E16` to `AAAAAAAA`)

serverless-mfa-api/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+WORKDIR /src
+
+RUN mkdir override
+RUN mkdir server
+ENV BRANCH=develop
+
+RUN curl -sSfL https://raw.githubusercontent.com/cosmtrek/air/master/install.sh | sh -s -- -b $(go env GOPATH)/bin
+
+# Download the necessary serverless-mfa-api-go files for running the api in a local container
```
forevermatt, line 15: Would it be better to have our serverless-mfa-api-go repo publish itself as a Docker image, for this sort of test/dev use? That feels like something that would better belong in that repo rather than this one.

That would also remove the need for having any serverless-mfa-api-go code here in this otherwise PHP repo, which could be helpful for PR reviews on this repo (i.e. not requiring the reviewers to understand both Go and PHP).

forevermatt (commented):
Suggested some changes

application/common/components/MfaBackendManager.php
Baggerone: I don't think so.  This is a different class than the Mfa class. It doesn't have a type attribute.

serverless-mfa-api/Dockerfile
Baggerone: It was my semi-sneaky way of providing you with more go exposure. ;-)

codeship-services.yml
Baggerone: They're real values.

application/frontend/config/main.php
Baggerone: Not the way I've coded it.  At this point, `credId` is the same as the `key_handle_hash` that the mfa-api needs.

Should we consider making it related to the database record id?

forevermatt: I think it would be better to use the database record ID, like we're doing for the `mfaId`.

codeship-services.yml
forevermatt: If they're real values, do we need to be concerned about them being made public here? (Or by "real values" do you something other than "actual values from some production system"?)

serverless-mfa-api/Dockerfile
forevermatt: 😆 Okay, if that's the best reason available, then it _really_ seems better to ~~delete all this Go code~~ (😉) move this into the serverless-mfa-api-go repo, since that really is the right place for serverless-mfa-api-go code.

codeship-services.yml
Baggerone: They're real but not production system values.  They need to "work" with the api backend.

serverless-mfa-api/Dockerfile
Baggerone: I'll look into using the aws php sdk for initializing dynamodb data.  I had been using just the cli commands in a bash script, but they didn't seem to work right when I tried to add a byte array as the value of an entry.

application/common/models/Mfa.php
```diff
@@ -25,6 +25,8 @@ class Mfa extends MfaBase
     const EVENT_TYPE_VERIFY = 'verify_mfa';
     const EVENT_TYPE_DELETE = 'delete_mfa';
 
+    const DEFAULT_WEBAUTHN_LABEL = 'My webauthn key';
```
forevermatt, line 28: I think sticking with "Security Key" as the default label might be better, as most users wont' have any idea what "webauthn" means.

application/features/bootstrap/MfaContext.php
```diff
@@ -5,8 +5,10 @@
 use common\models\EmailLog;
 use common\models\Mfa;
 use common\models\MfaBackupcode;
+use common\models\MfaWebauthn;
 use common\models\User;
 use Webmozart\Assert\Assert;
+use yii\base\BaseObject;
```
forevermatt, line 11: `BaseObject` doesn't seem to be used in this file. Is this `use` statement necessary?

dynamorestart/DynamoRestart.php
```diff
@@ -0,0 +1,131 @@
...
+                'encryptedAppId' => [
+                    'S' => 'SomeEncryptedAppId',
+                ],
+                'encryptedKeyHandle' => [
+                    'S' => 'SomeEncryptedKeyHandle',
+                ],
+//                'encryptedCredentials' => [
+            //  This doesn't work. What binary representation would work here?
+//                    'B' => 'ed634d2c138412f0e5d0f85ac8bceac9264df24a0bf597e75038caf9bb7cb6363beb7b8e9c660475b730fa4f29222b481cc76231d79ea8f8e8a4b0b2ebca3c315e9309db62c07ef0d4264073f1f6741b600086af6fa2d8657f660a1d415fc65ac907e2828865940fe2bfcc977577df1b35463dd04432dc2a746ca712e326ede06e3fa9d72f0a274d',
+//                ],
```
forevermatt, line 111: You might be able to use `hex2bin()`, though I don't know for sure:
```suggestion
                'encryptedCredentials' => [
                    'B' => hex2bin('ed634d2c138412f0e5d0f85ac8bceac9264df24a0bf597e75038caf9bb7cb6363beb7b8e9c660475b730fa4f29222b481cc76231d79ea8f8e8a4b0b2ebca3c315e9309db62c07ef0d4264073f1f6741b600086af6fa2d8657f660a1d415fc65ac907e2828865940fe2bfcc977577df1b35463dd04432dc2a746ca712e326ede06e3fa9d72f0a274d'),
                ],
```

forevermatt (approved):
A few more comments, but I _think_ you're good to go now (with or without the suggested changes).

dynamorestart/DynamoRestart.php
Baggerone: Thanks. That worked!

application/features/bootstrap/MfaContext.php
```diff
@@ -25,13 +32,16 @@ class MfaContext extends \FeatureContext
      */
     public function iGiveThatUserAVerifiedMfa($mfaType)
     {
+        Assert::notEq($mfaType, mfa::TYPE_WEBAUTHN, "should have called iGiveThatUserAVerifiedWebauthnMfa");
```
Schparky, line 35: Is this still needed? Looks like a test troubleshooting line.

application/features/mfa.feature
```diff
@@ -61,7 +61,6 @@ Feature: MFA
     Then the response status code should be 400
 
   Scenario: Create new MFA record of type webauthn
-#TODO - create a test double for the webauthn client
```
Schparky, line 64: Does this TODO still need to be here, perhaps with a different comment, since the scenario isn't implemented?

application/features/mfa.feature
```diff
@@ -120,3 +119,29 @@ Feature: MFA
...
+  Scenario: Delete the legacy u2f credential of a webauthn MFA option
+    Given the user has a verified webauthn MFA with a key_handle_hash of "u2f"
+    # This value is from a serverless-mfa-api-go test user that has a credential id of "C10"
+    # which gets hashed and base64 encoded to provide the "kI1ykA4kdZIbWA6XHmA-8iTxmVzfR-MCLRLuiK4-boo" value
+      And the user has a verified webauthn MFA with a key_handle_hash of "kI1ykA4kdZIbWA6XHmA-8iTxmVzfR-MCLRLuiK4-boo"
+    When I request to delete the webauthn entry of the MFA
+    Then the response status code should be 204
+      And the MFA record is still stored
+
+  Scenario: Delete the legacy u2f credential of a webauthn MFA option
```
Schparky, line 136: The scenario description doesn't seem to match. Isn't it deleting a webauthn credential?

application/features/mfa.feature
```diff
@@ -120,3 +119,29 @@ Feature: MFA
       And 0 codes should be stored
       And the MFA record is not stored
 
+  Scenario: Exception to delete the missing credential of a webauthn MFA option
+    Given the user has a verified webauthn MFA with a key_handle_hash of "u2f"
+    When I request to delete the webauthn entry of the MFA with a webauthn_id of 999
+    Then the response status code should be 404
+
+  Scenario: Delete the legacy u2f credential of a webauthn MFA option
```
Schparky, line 127: The scenario description doesn't seem to match. Isn't it deleting a webauthn credential?

application/frontend/config/main.php
```diff
@@ -60,11 +60,12 @@
...
-                'POST   mfa'                          => 'mfa/create',
-                'PUT    mfa/<id:\d+>'                 => 'mfa/update',
-                'POST   mfa/<id:\d+>/verify'          => 'mfa/verify',
-                'DELETE mfa/<id:\d+>'                 => 'mfa/delete',
+                'GET    user/<employeeId:\w+>/mfa'              => 'mfa/list',
+                'POST   mfa'                                    => 'mfa/create',
+                'PUT    mfa/<id:\d+>'                           => 'mfa/update',
+                'POST   mfa/<id:\d+>/verify'                    => 'mfa/verify',
+                'DELETE mfa/<id:\d+>'                           => 'mfa/delete',
+                'DELETE mfa/<mfaId:\d+>/webauthn/<webauthnId:\d+>' => 'mfa/delete-credential',
```
Schparky, line 68: I don't see a change to the `raml` file to correspond with this new endpoint.

Schparky (approved):
Good work here. I added some comments, some of which should be addressed before merging, if possible.

application/common/models/Mfa.php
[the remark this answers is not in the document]
Baggerone: Good catch.  It's not actually used anymore.  I changed the code that was using it but forgot to delete it here.

Makefile
```diff
@@ -43,19 +43,17 @@ quicktest:
...
+	docker-compose up -d mfaapi
 
 clean:
 	docker-compose kill
 	docker system prune -f
 
-raml2html: api.html
-
-api.html: api.raml
+raml2html:
```
Schparky, line 56: It was done this way so Make could skip the build if no changes were found in the raml file. Pretty insignificant difference, so the change is fine.

docker-compose.yml
```diff
@@ -195,7 +195,7 @@ services:
         volumes:
             - ./api.raml:/api.raml
             - ./api.html:/api.html
-        command: -i api.raml -o api.html
+        command: mattjtodd/raml2html -i api.raml -o api.html
```
Schparky, line 198: Did you encounter a problem with this line? It works both ways for me.

Makefile
Baggerone: It wasn't building for me even though the api.html file was empty.

docker-compose.yml
Baggerone: I did this while I was troubleshooting.  I think it helps (for future developers) to make the command explicit.  It's strange to me that it works without it.
The docs say `command overrides the default command declared by the container image (i.e. by Dockerfile’s CMD).`

Schparky:
I believe it's the norm to only provide arguments to the `ENTRYPOINT` command. In this case, it is `ENTRYPOINT ["raml2html"]`. (https://www.cloudbees.com/blog/understanding-dockers-cmd-and-entrypoint-instructions)

Baggerone:
OK. I can revert it in my next PR, if that's OK?

PR #271, opened 2022-08-10
Revert "Break out a mfa_webauthn child table to allow for individually deleting webauthn credentials"

Baggerone:
Reverts silinternational/idp-id-broker#267, since it appears to be breaking appsdev-pw-iidp.net


Baggerone:
Not ready to do this.

PR #273, opened 2022-08-11
Verify webauthn registration and include webauthn labels in response from MFA list

Baggerone:


application/common/models/Mfa.php
```diff
@@ -80,6 +82,26 @@ public function loadData(string $rpOrigin = '')
...
+            return;
+        }
+
+        if (! isset($this->data)) {
+            $this->data = [];
+        }
+
+        $webauthns = MfaWebauthn::findAll(['mfa_id' => $this-> id]);
+        foreach ($webauthns as $webauthn) {
+            $this->data += ['id' => $webauthn->id, 'label' => $webauthn->label];
```
Schparky, line 101: I am not sure what this _should_ do, but it only keeps the first pair of `id` and `label` elements.

application/common/models/Mfa.php
```diff
@@ -205,14 +227,30 @@ protected function hasTooManyRecentFailures(): bool
...
+        if ($verifyType != "") {
+            if ($this->type != self::TYPE_WEBAUTHN) {
+                throw new BadRequestHttpException(
+                    'A non-blank verification type is not allowed when verifying a mfa of type ' . $this->type
+                );
+            }
+
+            if ($verifyType != Mfa::VERIFY_REGISTRATION) {
+                throw new BadRequestHttpException(
+                    'A non-blank verification type for a ' . self::TYPE_WEBAUTHN . " may only be: " . self::VERIFY_REGISTRATION
```
Schparky, line 249: I'm a bit uncomfortable with using a string like a boolean. I might suggest using an actual `bool` or using two separate, non-empty strings for the two values.

Schparky (approved):

application/common/models/Mfa.php
Baggerone: OK, I saw something similar with a `+=`, but I guess it does something different than what I was expecting.

How about 
`
            $this->data[] = ['id' => $webauthn->id, 'label' => $webauthn->label];`

application/common/models/Mfa.php
Baggerone: @forevermatt Do you have a preference?  I guess I could have `actionVerify` convert the path parameter to a boolean with something like `verifyingRegistration = ($type == "registration")`

Is there a better way to do it?

Baggerone: @forevermatt Do you have a preference? I guess I could have actionVerify convert the path parameter to a boolean with something like verifyingRegistration = ($type == "registration")

Is there a better way to do it? (Is there a way to have a proper boolean come through in a url path with Yii2 routes?)

application/common/components/MfaBackendBackupcode.php
```diff
@@ -60,10 +60,11 @@ public function authInit(int $mfaId, string $rpOrigin = ''): array
      * @param int $mfaId The MFA ID
      * @param string $value Value provided by user, such as TOTP number or WebAuthn challenge response
      * @param string $rpOrigin
+     * @param string $verifyType Only used for WebAuthn
```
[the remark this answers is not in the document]
Baggerone: @forevermatt 
What do you prefer?
Is there an acceptable way in PHP to use a small class that is just intended for passing parameter values?

application/frontend/controllers/MfaController.php
```diff
@@ -90,7 +96,7 @@ public function actionVerify(int $id)
 
         // Strip spaces from $value if string
         if (is_string($value)) {
-            $value = preg_replace('/\D/', '', $value);
+            $value = preg_replace('/\s+/', '', $value);
```
forevermatt, line 99: It may be worth adding a comment here, either saying what `\s` represents or merely pointing to the documentation:
https://www.php.net/manual/en/regexp.reference.character-classes.php

application/frontend/controllers/MfaController.php
```diff
@@ -57,13 +57,19 @@ public function actionCreate()
...
      * @return null
      */
-    public function actionVerify(int $id)
+    public function actionVerify(int $id, string $type = "")
     {
+
+        if ($type != "" && $type != Mfa::VERIFY_REGISTRATION) {
+            throw new BadRequestHttpException('Verify type can only be: ' . Mfa::VERIFY_REGISTRATION . " Got: $type." , 1660222364);
+        }
+
```
forevermatt, line 72: I think this check could probably be left to the Mfa->verify() or to the MfaBackend(Type)->verify() function, rather than putting it up here in the controller. Just a suggestion.

application/frontend/controllers/MfaController.php
```diff
@@ -179,12 +188,28 @@ protected function getRequestedMfa($id)
...
+    {
+        $mfa = $this->getAuthorizedMfa($id);
+
         // rpOrigin is needed for WebAuthn authentication
         $rpOrigin = \Yii::$app->request->get('rpOrigin', '');
         if ($rpOrigin != '' && !in_array($rpOrigin, \Yii::$app->params['authorizedRPOrigins'])){
             throw new ForbiddenHttpException("Invalid rpOrigin", 1638539680);
         }
         $mfa->loadData($rpOrigin);
+        $mfa->loadMfaWebauthns();
```
forevermatt, line 212: What would you think of moving the logic currently in `loadMfaWebauthns()` into the `loadData()` function? It seems like it's also loading extra information into the `data` field, just for a specific type of MFA, and `loadData()` already has some conditions related to loading different data for specific types of MFAs.

application/frontend/controllers/MfaController.php
```diff
@@ -152,7 +161,7 @@ public function actionList(string $employeeId): array
      * @throws BadRequestHttpException
      * @throws NotFoundHttpException
      */
-    protected function getRequestedMfa($id)
+    protected function getAuthorizedMfa($id)
```
forevermatt, line 164: What does "authorized" mean here? Is this just meant to return verified MFA records? Or something else? (Thanks)

application/frontend/controllers/MfaController.php
```diff
@@ -261,7 +286,7 @@ public function actionDeleteCredential(int $mfaId, int $webauthnId)
      */
     public function actionUpdate(int $id)
     {
-        $mfa = $this->getRequestedMfa($id);
```
forevermatt, line 264: Was this change intentional? I wasn't sure if it was just the result of an IDE-assisted rename, or if the purpose of the `getRequestedMfa()` function changed, and so a different function needed to be called in order to do what `actionUpdate()` was previously using `getRequestedMfa()` to do.

application/frontend/controllers/MfaController.php
```diff
@@ -282,4 +307,48 @@ public function actionUpdate(int $id)
...
+     */
+    public function actionUpdateWebauthn(int $mfaId, int $webauthnId)
+    {
+        // check authorization
+        $this->getAuthorizedMfa($mfaId);
+
+        $label = \Yii::$app->request->getBodyParam('label');
+        if (empty($label)) {
+            throw new BadRequestHttpException('"label" must have a value in calls to update-webauthn', 1660233368);
+        }
```
forevermatt, line 330: I think this "a label is required" check probably belongs in the MfaWebauthn model's `rules()` function, rather than here in the controller.

forevermatt (changes requested):
A number of suggestions, a few questions, and one or two changes that I think really need to be made before this is merged.

application/features/mfa.feature
```diff
@@ -60,8 +70,20 @@ Feature: MFA
     When I request "/mfa" be created
     Then the response status code should be 400
 
-  Scenario: Create new MFA record of type webauthn
-#TODO - create tests that make use of the webauthn client
+  Scenario: Request a MFA record of type webauthn
```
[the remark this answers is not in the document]
Baggerone: This test shows us that idp-id-broker is making a valid call to serverless-mfa-api.  
Also, if it gets a `200` back from mfa-api but there is no `publicKey` value, that could mean the test called the wrong endpoint.

application/frontend/controllers/MfaController.php
Baggerone: The comment was already there above the `if` statement.

application/common/components/MfaBackendWebAuthn.php
```diff
@@ -125,29 +127,34 @@ public function verify(int $mfaId, $value, string $rpOrigin = '')
...
-                $mfa->key_handle_hash = $results['key_handle_hash'];
-                if (! $mfa->save()) {
-                    throw new ServerErrorHttpException(
-                        "Unable to save WebAuthn record after verification. Error: " . print_r($mfa->getFirstErrors(), true)
-                    );
-                }
-                return true;
+        // If the verifyType is not registration, finish the login process.
+        if ($verifyType != Mfa::VERIFY_REGISTRATION) {
+            $webauthn = MfaWebauthn::findOne(['mfa_id' => $mfa->id,'verified' => 1]);
```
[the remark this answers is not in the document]
Baggerone: How about this ...
```
            // Ensure there is at least one verified MfaWebauthn
            $webauthnCount = MfaWebauthn::find()->where(['mfa_id' => $mfa->id,'verified' => 1])->count();
            if ($webauthnCount < 1) {
                throw new NotFoundHttpException("Verified MFA Webauthn record not found for MFA ID: " . $mfa->id, 1659637860);
            }
            ```

File: application/common/models/Mfa.php [hunk 12]
```diff
@@ -80,6 +82,26 @@ public function loadData(string $rpOrigin = '')
...
+            return;
+        }
+
+        if (! isset($this->data)) {
+            $this->data = [];
+        }
+
+        $webauthns = MfaWebauthn::findAll(['mfa_id' => $this-> id]);
+        foreach ($webauthns as $webauthn) {
+            $this->data[] = ['id' => $webauthn->id, 'label' => $webauthn->label];
```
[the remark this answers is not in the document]
Baggerone: Yep. I'll add `, 'verified' => 1`

application/features/mfa.feature
```diff
@@ -10,12 +10,22 @@ Feature: MFA
...
 
-  Scenario: Retrieve MFA records for a User with 1 MFA record
+  Scenario: Retrieve MFA records for a User with a backupcode MFA record
     Given the user has a verified "backupcode" MFA
     When I request "/user/123/mfa" be retrieved
     Then the response status code should be 200
       And I should receive 1 record
 
+  Scenario: Retrieve MFA records for a User with a MfaWebauthn record
+    Given the user has a verified mfaWebauthn with a key_handle_hash of "KHH"
```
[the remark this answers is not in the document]
Baggerone: The main benefits of leaving it in there is that it is easy to see the value (without searching for it) and makes it easy to understand what to put in the `Then` lines 
 ````
 And the following mfaWebauthn data should be stored:
        | property            | value           |
        | label               | Security Key    |
        | key_handle_hash     | KHH             |
        ```

application/frontend/controllers/MfaController.php
Baggerone: It also checks that the user is authorized to deal with that Mfa.
Let me know if you prefer a different name for it.

application/frontend/controllers/MfaController.php
Baggerone: `loadData` is also called by User::getVerifiedMfaOptions(), which doesn't call `loadMfaWebauthns`.
I don't know if it matters. Knowing that, do you have a preference?

application/frontend/controllers/MfaController.php
Baggerone: I suppose the name of `getRequestedMfa()` should change (or both of them).  It calls `getAuthorizedMfa()` and then also does some other things.

`actionUpdate` only needed the mfa without the other things loaded onto it.

Maybe we should name it `getAuthorizedMfaWithExtras()`.  If we change the name of `getAuthorizedMfa()`, then of course the name of this function could be made to correspond to that new name.

application/frontend/controllers/MfaController.php
forevermatt: Looks like this should be reverted, since `$value` apparently is only supposed to be a string for TOTP. For WebAuthn, `$value` should already be a key/value array.

application/frontend/controllers/MfaController.php
```diff
@@ -139,11 +140,14 @@ public function actionList(string $employeeId): array
         $mfaOptions = Mfa::findAll(['user_id' => $user->id, 'verified' => 1]);
         foreach ($mfaOptions as $opt) {
             $opt->loadData($rpOrigin);
+            $opt->loadMfaWebauthns();
```
forevermatt, line 143: As discussed, please combine these.

application/frontend/controllers/MfaController.php
forevermatt: Revert to `getRequestedMfa($id)` (per discussion)

application/frontend/controllers/MfaController.php
forevermatt: Yeah, it's fine for that to include the MFA WebAuthn data as well. (So merge it in with `loadData()`).

application/frontend/controllers/MfaController.php
```diff
@@ -200,7 +220,7 @@ protected function getRequestedMfa($id)
      */
     public function actionDelete(int $id)
     {
-        $mfa = $this->getRequestedMfa($id);
+        $mfa = $this->getAuthorizedMfaWithExtras($id);
```
forevermatt, line 223: ... and these will be reverted.

application/frontend/controllers/MfaController.php
forevermatt: Thanks for adding that to `rules()`. Remember to remove it from here in the Controller, too.

application/common/components/MfaBackendTotp.php
```diff
@@ -35,6 +36,44 @@ class MfaBackendTotp extends Component implements MfaBackendInterface
...
+     * @param string $value Value provided by user, such as TOTP number or WebAuthn challenge response
+     * @param string $rpOrigin
+     * @param string $verifyType Only used for WebAuthn
+     * @return bool
+     * @throws BadRequestHttpException
+     * @throws NotFoundHttpException
+     * @throws ServerErrorHttpException
+     * @throws \Exception
+     */
+    public function verify(int $mfaId, string $value, string $rpOrigin = '', string $verifyType = ''): bool
```
forevermatt, line 51: If you don't mind moving this back down (it was somehow accidentally moved above `init()`), that would make the diff easier to read.

application/common/components/MfaBackendWebAuthn.php
```diff
@@ -125,29 +135,65 @@ public function verify(int $mfaId, $value, string $rpOrigin = '')
...
+        }
+
+        // Assume a new webauthn was requested and finish its registration process
+        $results = $this->client->webauthnValidateRegistration($headers, $value);
+
+        if (! isset($results['key_handle_hash'])) {
             return false;
         }
+
+        $label = $mfa->getReadableType();
```
forevermatt, line 163: This can be removed.

application/common/models/Mfa.php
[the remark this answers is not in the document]
forevermatt: Per our discussion, it looks like `mfa_webauthn.verified` is an unnecessary field, since we don't create the record until after we verify the registration attempt.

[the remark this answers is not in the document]
forevermatt: When you move this loop up into `loadData()`, you will want to run `make basemodels` to regenerate our "...Base" model classes. Then there will be a `$mfa->mfaWebauthns` property that is a list of all of those child records (avoiding the need to manually call `findAll()`.

application/common/components/MfaBackendWebAuthn.php
```diff
@@ -125,29 +135,65 @@ public function verify(int $mfaId, $value, string $rpOrigin = '')
...
+
+
+
+    /**
+     * @param string $label
+     * @param string $keyHandleHash
+     * @return void
+     * @throws ServerErrorHttpException
+     */
+    public static function createWebauthn(Mfa $mfa, string $keyHandleHash): void {
```
forevermatt, line 177: Sorry, I was wrong about moving this `createWebauthn()` function down into this `MfaBackendWebAuthn` class. The creating of MfaWebauthn records really does belong in that MfaWebauthn class. This "`MfaBackendWebAuthn` is really for handling interactions with the MFA API that are related to webauthn.

application/common/components/MfaBackendWebAuthn.php
```diff
@@ -125,29 +135,65 @@ public function verify(int $mfaId, $value, string $rpOrigin = '')
...
+        if ($mfa->type != self::TYPE_WEBAUTHN) {
+            return;
+        }
```
forevermatt, line 180: Counter to some of my earlier suggestions (😉), you will want to keep this sanity check when you move this back to MfaWebauthn, since it will need to be public. However you could change this to throw an exception, so that if we do later accidentally call this in some non-WebAuthn context, we'll know about that mistake as soon as we make it.

application/features/bootstrap/FeatureContext.php
```diff
@@ -590,6 +612,24 @@ public function getResponseProperty($property)
         return $this->resBody[$property];
     }
 
+    /**
+     * @param $property
+     * @return mixed
+     */
+    public function setRequestBody(string $key, string $value)
```
forevermatt, line 619: You'll probably want to change this to allow `$value` to be anything, since it will automatically be JSON-encoded later and needs to still be an object at this point for WebAuthn values.
```suggestion
    public function setRequestBody(string $key, $value)
```

application/features/mfa.feature
```diff
@@ -60,8 +70,20 @@ Feature: MFA
...
-  Scenario: Create new MFA record of type webauthn
-#TODO - create tests that make use of the webauthn client
+  Scenario: Request a MFA record of type webauthn
+    When the user requests a new webauthn MFA
+    Then the response status code should be 200
+    And the response body should contain 'publicKey'
+    And the response body should contain 'challenge'
+
+  Scenario: Verify a new MFA webauthn registration
+    Given the user has requested a new webauthn MFA
```
forevermatt, line 80: Let's revise this to be clearly intended for testing invalid values, perhaps like this:
```suggestion
  Scenario: Attempt to verify a new but invalid MFA webauthn registration
    Given the user has requested a new webauthn MFA with all the required fields but invalid values
```

application/features/mfa.feature
```diff
@@ -60,8 +70,20 @@ Feature: MFA
...
+    When I request to verify the webauthn Mfa registration
+    # Normally this would return a 200. However, we're not using completely correct
+    # webauthn data, due to its complexity and lack of predictability.
+    Then the response status code should be 500
+    And the response body should contain '400 Bad Request'
+    And the response body should contain '"error":"unable to create credential: Error validating challenge"'
```
forevermatt, line 86:
```suggestion
    Then the response status code should be 400
    And the response body should contain '400 Bad Request'
    And the response body should contain '"error":"unable to create credential: Error validating challenge"'
```

application/features/bootstrap/MfaContext.php
```diff
@@ -126,6 +155,50 @@ public function theUserDoesNotHaveAManagerEmailAddress()
...
+        Assert::notEmpty($id, 'Unable to get id of new Webauthn MFA');
+        $mfa = Mfa::FindOne(['id'=>$id]);
+        Assert::notEmpty($mfa, 'Unable to find that MFA.');
+
+        // It is too complicated at this point to come up with completely correct values
+        // These should get as far as producing a 400 status code with
+        // "error":"unable to create credential: Error validating challenge"
+
+        // These values are from values produced in serverless-mfa-api-go/webauthn_test.go Test_FinishRegistration
+        $reqValue = [
```
forevermatt, line 178: Since this will be refactored to be an obviously invalid set of values (at least partially), if there's a way to make it clear when reading the code that some piece of it is invalid, that could be helpful. I'm not sure of a good way to do that, though.

application/features/bootstrap/MfaContext.php
```diff
@@ -126,6 +155,50 @@ public function theUserDoesNotHaveAManagerEmailAddress()
...
+            ],
+            'user' => [
+                'displayName' => $user->display_name,
+                'id' => $user->id,
+                'name' => $user->username,
+            ],
+            'transports' => ['usb'],
+        ];
+
+        $reqJson = json_encode($reqValue);
```
forevermatt, line 195: You'll want to remove this, too, as it should be the array being passed, not a JSON-encoded string of the array.
```suggestion
```

application/frontend/controllers/MfaController.php
```diff
@@ -90,7 +91,7 @@ public function actionVerify(int $id)
 
         // Strip spaces from $value if string
```
forevermatt, line 92: Let's also update this comment to clarify the purpose of this `is_string()` check (i.e. to see whether is TOTP, and thus a string of digits, or not, and thus could be anything and shouldn't have non-digits deleted).

forevermatt (changes requested):

application/features/mfa.feature
```diff
@@ -138,13 +136,12 @@ Feature: MFA
       | employee_id | 123          |
       | label       |              |
     When I update the mfaWebauthn
-    Then the response status code should be 400
-      And the response body should contain '"label" must have a value in calls to update-webauthn'
+    Then the response status code should be 500
```
forevermatt, line 139: I think this response status code should probably be 400, since it was a bad request (to update the mfa_webauthn record but providing no label).

application/frontend/controllers/MfaController.php
```diff
@@ -179,6 +183,10 @@ protected function getRequestedMfa($id)
             );
         }
 
+        if ($mfa->type != Mfa::TYPE_WEBAUTHN) {
+            return $mfa;
+        }
```
forevermatt, line 188: I don't think we can actually return yet for all non-WebAuthn MFA records, since the `loadData()` call below adds data for other types of MFA records as well. Was there a need to return here, or was it simply because the code below seemed unnecessary for non-WebAuthn MFA records (or for some other reason)?

application/features/mfa.feature
Baggerone: This came from moving the verification to the `rules` on the model.  What's the best way to do this?

application/frontend/controllers/MfaController.php
```diff
@@ -282,4 +290,45 @@ public function actionUpdate(int $id)
...
+
+        if ($webauthn->update() === false) {
+            \Yii::error([
+                'action' => 'update mfa webauthn',
+                'status' => 'error',
+                'error' => $webauthn->getFirstErrors(),
+                'mfa_id' => $webauthn->id,
+            ]);
+            throw new ServerErrorHttpException("Unable to update MfaWebauthn label", 1660233472);
+        }
```
forevermatt, line 330: This is probably where we'll either need to first call `$webauthn->validate()` and return a 400 response if it fails, or else just return a 400 response for all unsuccessful calls to `$webauthn->update()`. I'm not sure whether those will all be bad-request failures, though. Some could be server errors, as you have it coded here currently.

application/features/mfa.feature
forevermatt: I added a later comment suggesting a way to do this just now. :-)

application/common/components/MfaBackendWebAuthn.php
```diff
@@ -125,29 +136,37 @@ public function verify(int $mfaId, $value, string $rpOrigin = '')
...
+        } catch (GuzzleException $e) {
+            throw new HttpException($e->getCode(), $e->getMessage(), 1660660611);
+        }
+
+        if (! isset($results['key_handle_hash'])) {
             return false;
         }
+
+        MfaWebauthn::createWebauthn($mfa, $results['key_handle_hash']);
+        $mfa->setVerified();
```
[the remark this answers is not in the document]
Baggerone: I think it's probably OK as is, because we try to delete the parent Mfa when the last webauthn child is deleted.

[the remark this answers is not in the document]
forevermatt: Oh, true. Thanks. Good point.

forevermatt (approved):
🎉

PR #276, opened 2022-08-24
Upgrade to PHP 8.1

Baggerone:


application/composer.json
```diff
@@ -18,7 +18,7 @@
     "silinternational/php-env": "^2.1.1",
     "yiisoft/yii2": "^2.0.11",
     "yiisoft/yii2-gii": "*",
-    "silinternational/psr3-adapters": "^1.1",
+    "silinternational/psr3-adapters": "^3.1.0",
```
mtompset, line 21: Nice. :)

Schparky (approved):

jason-jackson (approved):

Makefile
```diff
@@ -13,6 +13,7 @@ deps:
 	docker-compose run --rm cli composer install
 
 depsfortests:
+	docker-compose run --rm appfortests composer update
```
forevermatt, line 16: I'm a little hesitant to ever automatically update dependencies, though it's also very tempting and could be a good choice. Was there a particular motivator for adding this here?

application/versions.json
```diff
@@ -2,98 +2,155 @@
     "installed": [
         {
             "name": "behat/behat",
-            "version": "v3.10.0",
-            "description": "Scenario-oriented BDD framework for PHP"
+            "homepage": "http://behat.org/",
+            "source": "https://github.com/Behat/Behat/tree/v3.11.0",
+            "version": "v3.11.0",
+            "description": "Scenario-oriented BDD framework for PHP",
+            "abandoned": false
```
forevermatt, line 9: These JSON blocks are getting to have so many fields that this diff is getting hard to read again.

Not necessary in this PR (unless you just want to), but it may be worth it for us to do something to simplify the contents of our `versions.json` file to make it easier (again) to see what the actual updates were to the direct dependencies we name in our composer.json file. Just a thought.

forevermatt (approved):
Approved, with a question about the `composer update` in "depsfortests"

Makefile
```diff
@@ -13,6 +13,7 @@ deps:
...
 	docker-compose run --rm appfortests composer install
 
 depsshow:
```
Schparky, line 19: Removing `-f json` from the `composer show` command (or changing to `-f text`) [doc](https://getcomposer.org/doc/03-cli.md#show) makes for a much simpler output. We made the decision to use json, but I don't recall what the reasons were. @forevermatt do you remember?

application/versions.json
Schparky: Agreed. I think it would be a simple change. As commented above, just remove  `-f json` from the command in the Makefile.

Schparky (approved):

Makefile
Schparky: Here's one bit of history: https://github.com/silinternational/ssp-base/pull/61

PR #279, opened 2022-09-22
Upgrade yii json log targets and don't run tests until containers get settled

Baggerone:


Makefile
```diff
@@ -42,6 +42,7 @@ quicktest:
 	docker-compose run --rm test bash -c "vendor/bin/behat --stop-on-failure --strict --append-snippets"
 
 test: appfortests
+	sleep 3 # let other containers get settled
```
Schparky, line 45: Can you use `whenavail` or at least include in your comment which container(s) are the problem?

Schparky (approved):

PR #282, opened 2022-12-14
Multi-webauthn functionality

Baggerone:


Schparky (approved):
Assuming this isn't a code review so much as an approval process to move this one step closer to release. As such, I will approve, knowing that you have a timeline in mind for synchronizing these changes with the other pieces of this enhancement.

forevermatt (commented):
Didn't get all the way through the PR review, but did make it part way through. Just comments, nothing critical.

forevermatt (commented):
I have made it about halfway through this PR now. I'll try to finish within the next few workdays... sorry it's taking so long.

application/frontend/controllers/MfaController.php
```diff
@@ -246,4 +286,53 @@ public function actionUpdate(int $id)
...
+     * @return MfaWebauthn
+     * @throws BadRequestHttpException
+     * @throws NotFoundHttpException
+     * @throws ServerErrorHttpException
+     * @throws \Throwable
+     * @throws \yii\db\StaleObjectException
+     */
+    public function actionUpdateWebauthn(int $mfaId, int $webauthnId)
+    {
+        $this->getRequestedMfa($mfaId);
```
forevermatt, line 303: Since the return value of this isn't used, it looks (at first glance) like this line might be unnecessary. It appears (when I look closer) that it's validating that the MFA records exists and has the given Employee ID.

Would you mind adding a comment to communicate those side benefits of calling this method, since it's not obvious from the function name itself?

dynamorestart/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM composer:latest
```
forevermatt, line 1: We may want to switch this from "latest" to "2" to avoid being surprised by possibly-breaking changes in Composer 3:
```suggestion
FROM composer:2
```
As with the others, this is just a suggestion.

forevermatt (approved):
Approved, with a few suggestions/questions. Thanks! 👍

forevermatt (approved):

PR #285, opened 2023-02-24
Fix bug with delete inactive users and test it

Baggerone:
This was the error we were getting  ...
```
"error": "SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`idbroker`.`mfa_webauthn`, CONSTRAINT `fk_mfa_webauthn_mfa_id` FOREIGN KEY (`mfa_id`) REFERENCES `mfa` (`id`) ON DELETE NO ACTION ON UPDATE NO ACTION
```

### DIFF
--- a/application/common/models/Mfa.php
+++ b/application/common/models/Mfa.php
@@ -148,7 +148,21 @@ class Mfa extends MfaBase
     public function beforeDelete()
     {
         $this->clearFailedAttempts('when deleting mfa record');
-        
+
+        // first delete the webauthn children to avoid a foreign key constraint error
+        foreach ($this->mfaWebauthns as $child) {
+            if (!$child->delete()) {
+                \Yii::error([
+                    'action' => 'delete mfa webauthn child record before deleting mfa',
+                    'status' => 'error',
+                    'error' => $child->getFirstErrors(),
+                    'mfa_id' => $this->id,
+                    'child_id' => $child->id,
+                ]);
+                return false;
+            }
+        }
+
         $backend = self::getBackendForType($this->type);
         return $backend->delete($this->id);
     }
--- a/application/common/models/User.php
+++ b/application/common/models/User.php
@@ -92,6 +92,13 @@ class User extends UserBase
         }
 
         foreach ($this->mfas as $mfa) {
+            // first delete the children and then the parent to avoid foreign key constraint error
+            if ($mfa->type == MFA::TYPE_WEBAUTHN) {
+                $didDelete = self::deleteWebauthns($mfa);
+                if (!$didDelete) {
+                    return false;
+                }
+            }
             if (! $mfa->delete()) {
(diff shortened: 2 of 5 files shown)


application/common/models/User.php
```diff
@@ -149,6 +155,26 @@ public function beforeDelete()
         return true;
     }
 
+    // The MfaBackend code for deleting a webauthn type MFA
```
Schparky, line 158: I don't understand this comment and why it's relevant. It sounds like justification for doing things the "hard" way or doing it here instead of in the MfaBackend code. But this implementation looks about as simple as it could be and doesn't look out of place to me.

Schparky (approved):

application/common/models/User.php
```diff
@@ -92,6 +92,12 @@ public function beforeDelete()
         }
 
         foreach ($this->mfas as $mfa) {
+            // first delete the children and then the parent to avoid foreign key constraint error
+            if ($mfa->type == MFA::TYPE_WEBAUTHN) {
+                if (!self::deleteWebauthns($mfa)) {
```
forevermatt, line 97: It will probably be cleaner to leave this `User->beforeDelete()` method unchanged, and instead update the Mfa record's `beforeDelete()` method to delete each of its MfaWebauthn children records when appropriate (using each one's `delete()` method).

application/common/models/User.php
```diff
@@ -149,6 +155,26 @@ public function beforeDelete()
...
+                \Yii::error([
+                    'action' => 'delete mfa webauthn child record before deleting mfa',
+                    'status' => 'error',
+                    'error' => $child->getFirstErrors(),
+                    'mfa_id' => $mfa->id,
+                    'child_id' => $child->id,
+                ]);
+                return false;
+            }
+        }
```
forevermatt, line 174: This is essentially the code that would be moved to the MfaWebauthn->beforeDelete() method, per my other comment.

forevermatt (changes requested):
Small request about where some of the new code is placed, but this is looking good other than that. 👍

application/common/models/Mfa.php
```diff
@@ -148,7 +148,21 @@ public function afterSave($insert, $changedAttributes)
     public function beforeDelete()
     {
         $this->clearFailedAttempts('when deleting mfa record');
-        
+
+        // first delete the webauthn children to avoid a foreign key constraint error
+        foreach ($this->mfaWebauthns as $child) {
```
Baggerone, line 153: I considered putting this in an if block that checked that the `mfa` was of type webauthn, but I figured none of the other types would have any mfaWebauthn children. So, the `foreach` loop would just be skipped anyway.

forevermatt (approved):
Looks good, thanks.

PR #291, opened 2023-03-20
Release 6.4.0

Baggerone:
## Changed
It now makes calls for Google Analytics 4. This includes replacing hyphens with underscores in the GA event labels/names, since hyphens are not accepted in GA4.

# Added
- `GaController` that provides a `cli` command which makes a call to GA4 for manual testing.
- Expects to have two new environment variables:
  -- GA_MEASUREMENT_ID
  -- GA_API_SECRET

# Removed
No longer expects the environment variable GA_TRACKING_ID


Schparky (approved):
By the same logic that made the change in idp-in-a-box a major change, this probably should be as well. But as I said over there, I can go either way on that decision.

Baggerone:
> By the same logic that made the change in idp-in-a-box a major change, this probably should be as well. But as I said over there, I can go either way on that decision.

I guess the difference I see is that a Tfm Cloud run fails when you include an undefined variable, whereas idp-id-broker as an app doesn't fail if you have the wrong GA env vars. (It just logs an error and continues.)

What would your preference for the tag be, now that I gave the Tfm module a major release?

Schparky:
> > By the same logic that made the change in idp-in-a-box a major change, this probably should be as well. But as I said over there, I can go either way on that decision.
> 
> I guess the difference I see is that a Tfm Cloud run fails when you include an undefined variable, whereas idp-id-broker as an app doesn't fail if you have the wrong GA env vars. (It just logs an error and continues.)

Ok, I was thinking that was not a critical error. In that case, it is definitely a breaking change on idp-in-a-box.

> What would your preference for the tag be, now that I gave the Tfm module a major release?

As you have it now is fine.

PR #292, opened 2023-04-21
IDP-26 re-use the same manager code 

Baggerone:
Am I right in thinking that this is the best way to deal with [IDP-26](https://itse.youtrack.cloud/issue/IDP-26) ?


application/features/mfa.feature
```diff
@@ -55,12 +55,21 @@ Feature: MFA
...
         | verified  | 1              |
+    When I request "/mfa" be created
+     And I provide the following valid data:
+       | property    | value          |
+       | employee_id | 123            |
+       | type        | manager        |
+       | label       | A Label        |
+    Then the response status code should be 200
+     And the response should contain a data array with 0 items
+     And the same MFA record exists with the same backup code
```
forevermatt, line 72: A Behat scenario is intended to only have a single "When" (coming after the Given section, and following by the Then section).

Would you mind breaking this "only once" test out into its own scenario?

forevermatt: It's also probably only important that the backup code itself is unchanged. I wouldn't compare the MFA records' IDs, since that seems like an implementation detail. If we later change how we implement this in such a way that a new record is created that has the same value, that still accomplishes the goal (and so, I think, shouldn't cause a test to fail).

forevermatt (changes requested):
Yes, this seems like a good way to do this. Requesting minor changes, but I like your direction with this.

Schparky (approved):

PR #299, opened 2023-06-27
Ensure required environment variables (as arrays) are present

Baggerone:
This keeps  AWS from killing the old task by not allowing the new task to become healthy and it creates the following logs in Cloudwatch.

```
{
    "name": "Internal Server Error",
    "message": "The MFA_TOTP_apiSecret environment variable is required",
    "status": 500
}
```

### DIFF
--- a/application/common/config/main.php
+++ b/application/common/config/main.php
@@ -9,6 +9,7 @@ use common\components\MfaBackendWebAuthn;
 use Sil\JsonLog\target\EmailServiceTarget;
 use Sil\JsonLog\target\JsonStreamTarget;
 use Sil\PhpEnv\Env;
+use Sil\PhpEnv\EnvVarNotFoundException;
 use yii\db\Connection;
 use yii\helpers\ArrayHelper;
 use yii\helpers\Json;
@@ -24,10 +25,25 @@ $notificationEmail = Env::get('NOTIFICATION_EMAIL');
 
 $mfaNumBackupCodes = Env::get('MFA_NUM_BACKUPCODES', 10);
 
+$envValidator = function($name, $key, $config) {
+    if (! isset($config[$key])) {
+        $message = "The $name environment variable is required";
+        throw new EnvVarNotFoundException($message);
+    }
+};
+
 $mfaTotpConfig = Env::getArrayFromPrefix('MFA_TOTP_');
 $mfaTotpConfig['issuer'] = $idpDisplayName;
 
+$envValidator('MFA_TOTP_apiBaseUrl', 'apiBaseUrl', $mfaTotpConfig);
+$envValidator('MFA_TOTP_apiKey', 'apiKey', $mfaTotpConfig);
+$envValidator('MFA_TOTP_apiSecret', 'apiSecret', $mfaTotpConfig);
+
 $mfaWebAuthnConfig = Env::getArrayFromPrefix('MFA_WEBAUTHN_');
+
+$envValidator('MFA_WEBAUTHN_apiBaseUrl', 'apiBaseUrl', $mfaTotpConfig);
+$envValidator('MFA_WEBAUTHN_apiKey', 'apiKey', $mfaTotpConfig);
+$envValidator('MFA_WEBAUTHN_apiSecret', 'apiSecret', $mfaTotpConfig);
 
 $emailerClass = Env::get('EMAILER_CLASS', Emailer::class);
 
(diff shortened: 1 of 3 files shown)


application/common/config/main.php
```diff
@@ -24,11 +25,26 @@
 
 $mfaNumBackupCodes = Env::get('MFA_NUM_BACKUPCODES', 10);
 
+$envValidator = function($name, $key, $config) {
```
forevermatt, line 28: Would you mind switching to simply declaring the function, rather than assigning the function to a variable?

```suggestion
function assertEnvVarIsSet($name, $key, $config) {
```

I think that will be easier to follow. In PHP, it's still (relatively) rare to see a function assigned to a variable (unlike in JavaScript). It's supported, just a little unexpected, and in this case it's not necessary to do so.

application/common/config/main.php
```diff
@@ -24,11 +25,26 @@
...
+        throw new EnvVarNotFoundException($message);
+    }
+};
+
 $mfaTotpConfig = Env::getArrayFromPrefix('MFA_TOTP_');
 $mfaTotpConfig['issuer'] = $idpDisplayName;
 
+$envValidator('MFA_TOTP_apiBaseUrl', 'apiBaseUrl', $mfaTotpConfig);
+$envValidator('MFA_TOTP_apiKey', 'apiKey', $mfaTotpConfig);
+$envValidator('MFA_TOTP_apiSecret', 'apiSecret', $mfaTotpConfig);
```
forevermatt, line 40:
```suggestion
assertEnvVarIsSet('MFA_TOTP_apiBaseUrl', 'apiBaseUrl', $mfaTotpConfig);
assertEnvVarIsSet('MFA_TOTP_apiKey', 'apiKey', $mfaTotpConfig);
assertEnvVarIsSet('MFA_TOTP_apiSecret', 'apiSecret', $mfaTotpConfig);
```

application/common/config/main.php
```diff
@@ -24,11 +25,26 @@
...
+
 $mfaWebAuthnConfig = Env::getArrayFromPrefix('MFA_WEBAUTHN_');
 
+$envValidator('MFA_WEBAUTHN_apiBaseUrl', 'apiBaseUrl', $mfaTotpConfig);
+$envValidator('MFA_WEBAUTHN_apiKey', 'apiKey', $mfaTotpConfig);
+$envValidator('MFA_WEBAUTHN_apiSecret', 'apiSecret', $mfaTotpConfig);
```
forevermatt, line 46:
```suggestion
assertEnvVarIsSet('MFA_WEBAUTHN_apiBaseUrl', 'apiBaseUrl', $mfaTotpConfig);
assertEnvVarIsSet('MFA_WEBAUTHN_apiKey', 'apiKey', $mfaTotpConfig);
assertEnvVarIsSet('MFA_WEBAUTHN_apiSecret', 'apiSecret', $mfaTotpConfig);
```

forevermatt (commented):
Requested stylistic change. You're welcome to name the function whatever you want; I just used something as an example.

application/common/config/main.php
```diff
@@ -24,11 +25,26 @@
 
 $mfaNumBackupCodes = Env::get('MFA_NUM_BACKUPCODES', 10);
 
+$envValidator = function($name, $key, $config) {
+    if (! isset($config[$key])) {
+        $message = "The $name environment variable is required";
+        throw new EnvVarNotFoundException($message);
+    }
+};
```
jason-jackson, line 33: If you wanted to, you could also make a function to load and validate at the same time, something like:
```suggestion
function loadConfig($name, $requiredKeys) {
    $config = Env::getArrayFromPrefix($name);
    foreach ($requiredKeys as $key) {
        if (! isset($config[$key])) {
            $message = "The {$name}{$key} environment variable is required";
            throw new EnvVarNotFoundException($message);
        }
    }
    return $config
};
```

forevermatt (approved):

application/common/config/main.php
Baggerone: (Just as a record) It looks like it can't use a normal function, since the file must be required by multiple other files, which makes it look like the function is being re-defined.

application/common/config/main.php
Baggerone: Thanks. I think the way I have it is easier to follow and may even take fewer lines in the end.

PR #302, opened 2023-06-30
require MFA config variables on status check

briskt:
### Fixed
- Check for required MFA configuration during the status check, not in root config file. This alleviates the need to provide dummy variables everywhere the broker image is used.
- Use a more deterministic dependency list format.
- Dependency update, including new `email-service-php-client` that reports DNS error correctly.
- Delete unused code (`Emailer->getSiteStatus()`)

### DIFF
--- a/application/frontend/controllers/SiteController.php
+++ b/application/frontend/controllers/SiteController.php
@@ -66,39 +66,6 @@ class SiteController extends BaseRestController
         Yii::$app->response->statusCode = 204;
     }
 
-    private function checkEmailServiceStatus($emailer)
-    {
-        try {
-            $emailer->getSiteStatus();
-        } catch (GuzzleCommandException $e) {
-            $response = $e->getResponse();
-            if ($response) {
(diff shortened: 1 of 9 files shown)


Baggerone:
I think I'll leave this for @forevermatt  to review. It doesn't seem right to me that something only AWS cares about would be the thing that checks whether all the required variables are present.

application/frontend/controllers/SiteController.php
forevermatt, line 1: Should/could the `checkEmailServiceStatus()` function also be removed?

briskt: Yes, thanks.

briskt:
> I'm willing to try this, though I don't see a strong reason either direction for whether to have the error happen when trying to load the required env. vars. or when checking site status... it looks like we're already doing a mix of those two approaches.

I had two reasons for doing this, one of which is included in the PR description. The other is that to me it seems better to check for required variables closer to where they are used.

PR #306, opened 2023-10-09
set YII_ENV and YII_DEBUG in yii console

briskt:
[IDP-698](https://itse.youtrack.cloud/issue/IDP-698)

### DIFF
--- a/application/yii
+++ b/application/yii
@@ -8,8 +8,12 @@
  * @license http://www.yiiframework.com/license/
  */
 
-defined('YII_DEBUG') or define('YII_DEBUG', true);
-defined('YII_ENV') or define('YII_ENV', 'dev');
+use Sil\PhpEnv\Env;
+
+require(__DIR__ . '/vendor/autoload.php');
+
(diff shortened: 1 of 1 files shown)


application/yii
```diff
@@ -8,8 +8,12 @@
  * @license http://www.yiiframework.com/license/
  */
 
-defined('YII_DEBUG') or define('YII_DEBUG', true);
-defined('YII_ENV') or define('YII_ENV', 'dev');
+use Sil\PhpEnv\Env;
+
+require(__DIR__ . '/vendor/autoload.php');
```
forevermatt, line 13: The composer autoload file is already being required below (new line 22). Can it be removed here? (or there)

briskt: Good observation. I suppose it would be better to remove the later one, but I'm not as comfortable with PHP includes as I used to be.

forevermatt: You'll need it included before you use the `Env` class, so either just keep the top one, or use PHP's built-in functions for getting the env. var. and keep the lower require. My suggestion.

PR #309, opened 2023-10-11
update the last_used_utc on the webauthn record IDP-690

briskt:
### Fixed
- Update the `last_used_utc` column on a webauthn record when it is used. [IDP-690](https://itse.youtrack.cloud/issue/IDP-690)
- Removed unused test code.

---

### Feature PR Checklist
- [x] Run `make composershow`

_Code and test structure does not lend itself to testing the verification of a webauthn. It may be possible using the U2F simulator. I am not familiar with how to use that, though._

### DIFF
--- a/application/common/components/MfaBackendWebAuthn.php
+++ b/application/common/components/MfaBackendWebAuthn.php
@@ -158,6 +158,9 @@ class MfaBackendWebAuthn extends Component implements MfaBackendInterface
             if ($webauthnCount < 1) {
                 throw new NotFoundHttpException("No MFA Webauthn record found for MFA ID: " . $mfa->id, 1659637860);
             }
+
+            $mfa->setLastUsed();
+
             return $this->client->webauthnValidateAuthentication($headers, $value);
         }
 
--- a/application/common/models/MfaWebauthn.php
+++ b/application/common/models/MfaWebauthn.php
@@ -92,4 +92,17 @@ class MfaWebauthn extends MfaWebauthnBase
 
         return $webauthn;
     }
+
+    public function setLastUsed()
+    {
+        $this->last_used_utc = MySqlDateTime::now();
+        if (! $this->save()) {
+            \Yii::error([
+                'action' => 'update webauthn last_used_utc',
+                'status' => 'error',
+                'mfa_id' => $this->id,
+                'error' => $this->getFirstErrors(),
+            ]);
+        }
+    }
 }
--- a/application/common/models/MfaWebauthn.php
+++ b/application/common/models/MfaWebauthn.php
@@ -96,7 +96,7 @@ class MfaWebauthn extends MfaWebauthnBase
     public function setLastUsed()
     {
         $this->last_used_utc = MySqlDateTime::now();
-        if (! $this->save()) {
+        if (! $this->save(true, ['last_used_utc'])) {
             \Yii::error([
                 'action' => 'update webauthn last_used_utc',
                 'status' => 'error',
--- a/application/features/bootstrap/MfaContext.php
+++ b/application/features/bootstrap/MfaContext.php
@@ -49,7 +49,7 @@ class MfaContext extends \FeatureContext
         ]);
 
         Assert::true($this->mfa->save(), 'Failed to add that MFA record to the database.');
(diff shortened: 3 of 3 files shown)


application/features/bootstrap/MfaContext.php
```diff
@@ -262,14 +262,6 @@ public function iRequestToVerifyTheWebauthnMfaRegistrationWithALabelOf($label)
         $this->iRequestTheResourceBe('/mfa/' . $this->mfa->id . '/verify/registration', 'created');
     }
 
-    /**
```
mtompset, line 265: Why was this removed?

briskt: I was trying to write a test that would use this, but failed. In the process, I noticed that it was not being used.

jason-jackson:
Looks like you should be able to use the u2fsimulator, it's using the `silintl/serverless-mfa-api-go:develop` 

Example usages:
```
$this->callU2fSimulator('/webauthn/login', 'created', $user, $mfa->external_uuid);
$this->callU2fSimulator('/webauthn/login', 'updated', $user, $mfa->external_uuid);

$this->callU2fSimulator('/webauthn/register', 'created', $user, $mfa->external_uuid);
$this->callU2fSimulator('/webauthn/register', 'updated', $user, $mfa->external_uuid);

$this->callU2fSimulator('/webauthn/user', 'deleted', $user, $mfa->external_uuid);
$this->callU2fSimulator('/webauthn/credential/'.$mfa->id, 'deleted', $user, $mfa->external_uuid);  // I could be wrong about this one
```

I'd have to look at proper order for calling as it's been a while since I worked with webauthn.

briskt:
Thanks, @jason-jackson. That's a step in the right direction. What I needed was the valid data that needs to be passed to the id-broker in an /mfa/verify call.

application/common/models/MfaWebauthn.php
```diff
@@ -92,4 +92,17 @@ public static function createWebauthn(Mfa $mfa, string $keyHandleHash, string $l
 
         return $webauthn;
     }
+
+    public function setLastUsed()
+    {
+        $this->last_used_utc = MySqlDateTime::now();
+        if (! $this->save()) {
```
mtompset, line 99: This saves more than just last_used_utc. You could limit it to just last_used_utc. But this works.

mtompset (approved):
I'm assuming someone has tested that it runs, because it eyeballs fine.

briskt:
> I'm assuming someone has tested that it runs, because it eyeballs fine.

I was planning on testing it on the staging environment.

application/common/models/MfaWebauthn.php
briskt: What's the syntax for that? I could look it up, but I'm sure you can whip it right out.

mtompset: $this->save(true, ['last_used_utc']) -- still runs the validation, but only validate and save last_used_utc.

briskt: Thank you very much.

application/common/components/MfaBackendWebAuthn.php
```diff
@@ -158,6 +158,9 @@ public function verify(int $mfaId, $value, string $rpOrigin = '', string $verify
             if ($webauthnCount < 1) {
                 throw new NotFoundHttpException("No MFA Webauthn record found for MFA ID: " . $mfa->id, 1659637860);
             }
+
+            $mfa->setLastUsed();
```
forevermatt, line 162: It looks like this is recording a last-used date before verifying that the response was valid. Is that correct?

briskt: Yes, it does look like I got that wrong. Thanks for spotting it.

briskt: What's worse, as I got to looking at it further, I'm not sure it's going to work at all. I tried to call `setLastUsed` on an `Mfa` object, but I defined the function on `MfaWebauthn`. It appears that the validation is delegated to the MFA API, which doesn't give any indication as to which webauthn instance passes validation.

PR #315, opened 2023-11-08
rehash password if the hash cost is too low

briskt:
### Security
- Rehash a user's password during verify if the hash cost is too low.

[IDP-126](https://itse.youtrack.cloud/issue/IDP-126)

### DIFF
--- a/application/common/models/Password.php
+++ b/application/common/models/Password.php
@@ -21,6 +21,8 @@ class Password extends PasswordBase
 {
     public const SCENARIO_UPDATE_METADATA = 'update_metadata';
 
+    // hash algorithm passed to PHPs `password_hash` function -- if this is changed, the options
+    // parameter passed to any `password_` functions may need to be changed as well
     public const HASH_ALGORITHM = PASSWORD_BCRYPT;
 
     public const HASH_COST = 13;
--- a/application/common/models/Password.php
+++ b/application/common/models/Password.php
@@ -19,7 +19,11 @@ use yii\web\ConflictHttpException;
  */
 class Password extends PasswordBase
 {
-    const SCENARIO_UPDATE_METADATA = 'update_metadata';
+    public const SCENARIO_UPDATE_METADATA = 'update_metadata';
+
+    public const HASH_ALGORITHM = PASSWORD_BCRYPT;
+
+    public const HASH_COST = 13;
 
     public $password;
 
@@ -46,7 +50,7 @@ class Password extends PasswordBase
             ],
             [
                 'hash', 'default', 'value' => function () {
-                    return password_hash($this->password, PASSWORD_DEFAULT);
+                    return self::hashPassword($this->password);
                 },
             ],
             [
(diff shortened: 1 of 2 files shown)


application/common/models/Password.php
```diff
@@ -280,4 +284,9 @@ public function markPwned(): void
             ]);
         }
     }
+
+    public static function hashPassword(string $password): string
+    {
+        return password_hash($password, self::HASH_ALGORITHM, ["cost" => self::HASH_COST]);
```
mtompset, line 290: Is this because the PHP Documentation says, "Note that this constant is designed to change over time as new and stronger algorithms are added to PHP."?

mtompset (approved):
My comment is not meant to hold up approval, but rather confirm my desire-stability assumption.

application/common/models/Password.php
```diff
@@ -19,7 +19,11 @@
  */
 class Password extends PasswordBase
 {
-    const SCENARIO_UPDATE_METADATA = 'update_metadata';
+    public const SCENARIO_UPDATE_METADATA = 'update_metadata';
+
+    public const HASH_ALGORITHM = PASSWORD_BCRYPT;
```
forevermatt, line 24: It may be worth adding a comment here that, if we ever change this to something other than `PASSWORD_BCRYPT`, we will need to change (potentially) all of the places in this file where we use PHP's built-in `password_*()` functions, because the options are different for different hashing algorithms (as you probably figured out before I did).

application/common/models/User.php
```diff
@@ -298,6 +298,15 @@ private function validatePassword(): Closure
...
             if (! password_verify($this->password, $currentPassword->hash)) {
                 $this->addError($attributeName, 'Incorrect password.');
+            } else {
+                // check the current hash cost and rehash if necessary
+                if (password_needs_rehash(
+                    $currentPassword->hash,
+                    Password::HASH_ALGORITHM,
+                    ['cost' => Password::HASH_COST]
+                )) {
+                    $this->savePassword();
```
forevermatt, line 308: It looks like this will create a new password record for the User, but won't set it as the current password for that user (unless there is a later call to User->save() that's not obvious from this code, which is possible).

forevermatt, line 308: Another interesting detail of this is that it will appear that the user changed their password anytime we do this, when in fact we simply re-hashed their existing password.

Is there a way would could instead change the hash on their existing password record? (Feel free to suggest alternatives.)

forevermatt (changes requested):
Overall, this looks good. I just have some concerns about the way we're saving the new password hash to the database.

application/common/models/Password.php
briskt: Ick. I'll make another pass at this.

application/common/models/Password.php
briskt: Not necessarily, I just thought it seemed like a better option than "default".

forevermatt (commented):
Thanks for the revision. Looks better. 👍

briskt:
> Thanks for the revision. Looks better. 👍

Thanks. Not sure if you noticed, but it logs an error even though it worked correctly. Suspicious.

briskt:
> > Thanks for the revision. Looks better. 👍
> 
> Thanks. Not sure if you noticed, but it logs an error even though it worked correctly. Suspicious.

Hopefully this helps: https://github.com/silinternational/idp-id-broker/commit/6403f5650e812a185c6b206269fc8c25ca1da0b6

BTW, @forevermatt, this PR is still listed as "Changes requested". I would have thought that gets cleared when merged.

PR #316, opened 2023-11-08
update php-cs-fixer and run "make psr2"

briskt:
### Fixed
- Updated php-cs-fixer to 3.x so it can run on the current version of PHP.
- Used `make psr2` to update PHP source files to consistent style.

---

### Feature PR Checklist
- [x] Run `make composershow`
- [x] Run `make psr2`


application/common/mail/mfa-option-added.html.php
```diff
@@ -32,7 +32,7 @@
         foreach ($mfaOptions as $mfa) {
             echo '<li>' . yHtml::encode($mfa->getReadableType()) . '</li>';
         }
-    ?>
+?>
```
forevermatt, line 35: Oh that's unfortunate (though minimal, all things considered).

forevermatt (approved):
Great, thanks!

PR #317, opened 2023-11-09
fix the password rehash feature

briskt:
### Fixed
- Fix the password rehash feature. It was failing with error "May not be reused yet".

---

### Feature PR Checklist
- [x] Unit tests created or updated
- [x] Run `make psr2`

### DIFF
--- a/application/common/models/Password.php
+++ b/application/common/models/Password.php
@@ -21,6 +21,8 @@ class Password extends PasswordBase
 {
     public const SCENARIO_UPDATE_METADATA = 'update_metadata';
 
+    public const SCENARIO_REHASH = 'rehash';
+
     // hash algorithm passed to PHPs `password_hash` function -- if this is changed, the options
     // parameter passed to any `password_` functions may need to be changed as well
     public const HASH_ALGORITHM = PASSWORD_BCRYPT;
@@ -48,7 +50,7 @@ class Password extends PasswordBase
                 'password', 'string',
             ],
             [
-                'password', 'checkRecentlyUsed',
+                'password', 'checkRecentlyUsed', 'on' => self::SCENARIO_DEFAULT
             ],
             [
                 'hash', 'default', 'value' => function () {
@@ -241,6 +243,8 @@ class Password extends PasswordBase
             'expires_on',
             'grace_period_ends_on',
         ];
+
+        $scenarios[self::SCENARIO_REHASH] = ['hash'];
 
         return $scenarios;
     }
--- a/application/common/models/User.php
+++ b/application/common/models/User.php
@@ -306,7 +306,7 @@ class User extends UserBase
                     ['cost' => Password::HASH_COST]
                 )) {
                     $currentPassword->hash = Password::hashPassword($this->password);
-                    $currentPassword->scenario = Password::SCENARIO_REHASH;
+                    $currentPassword->setScenario(Password::SCENARIO_REHASH);
                     if (!$currentPassword->save()) {
                         Yii::error([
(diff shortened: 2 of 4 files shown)


application/features/password.feature
```diff
@@ -3,18 +3,17 @@ Feature: Password
...
-#    Given I receive an existing employee id
-#      And the requestor is authorized
-#      And the user does not already have a password
-#      And I receive a password
-#    When I receive a request to create a password for a specific user
-#    Then a new password hash should be stored
-#      And the last changed date should be stored as the instant it was stored
-#      And the last changed date should be stored in UTC
-#      And the last synched date should be stored as the instant it was stored
-#      And the last synched date should be stored in UTC
```
jason-jackson, line 16: I think this is a decent test to make sure last changed and synced is updated with a new password as I think we use it to determine if a password needs to be changed. Or am I thinking of a different part of the IdP?

application/common/models/Password.php
```diff
@@ -48,7 +50,7 @@ public function rules(): array
                 'password', 'string',
             ],
             [
-                'password', 'checkRecentlyUsed',
+                'password', 'checkRecentlyUsed', 'on' => self::SCENARIO_DEFAULT
```
mtompset, line 53: Sorry, I am getting Yii 2.x upgrade triggered. You should probably figure out "new Password(['scenario' => 'update']);" or create instead of update. "on default" seems pointless?

application/common/models/User.php
```diff
@@ -305,7 +305,8 @@ private function validatePassword(): Closure
                     Password::HASH_ALGORITHM,
                     ['cost' => Password::HASH_COST]
                 )) {
-                    $currentPassword->password = $this->password;
+                    $currentPassword->hash = Password::hashPassword($this->password);
+                    $currentPassword->scenario = Password::SCENARIO_REHASH;
```
mtompset, line 309: I would use
$currentPassword->setScenario(Password::SCENARIO_REHASH);
Direct access of magical innards scares me. :)

application/features/password.feature
```diff
@@ -3,18 +3,17 @@ Feature: Password
   As an authorized user
   I need to be able to update a specific user password
 
-#  Scenario: Provide a new password for an existing user
-#    Given I receive an existing employee id
```
mtompset, line 7: Where are the functions for this? Should they not be removed?

application/features/password.feature
briskt: I think this `password.feature` was a placeholder written well before the system was fully complete. Or at least there are features listed here that are not in the current system. Specifically, the password table doesn't have a "last synced date" nor a "last changed date" column.

application/common/models/Password.php
briskt: Not sure I understand your comment fully, but one of the suggestions on an earlier PR was to reuse the existing Password record so the metadata doesn't change.

application/common/models/User.php
briskt: I was wondering about that. I saw `setScenario` in the Yii API late in my work here, and did a quick search on this repo. It seemed that a direct assignment to `scenario` is the "norm", but that doesn't make it right.

application/features/password.feature
briskt: I removed this for two reasons: it references features that do not exist in id-broker, and it is written like an integration test. In my opinion, unit tests have a higher priority than integration tests, which means that if you have neither, the unit tests should be written first. I would be glad to share more about that if you're interested.

PR #318, opened 2023-11-15
Rehash passwords with a higher "cost"

briskt:
### Fixed
- Updated php-cs-fixer to 3.x so it can run on the current version of PHP.
- Used `make psr2` to update PHP source files to consistent style.

### Security
- Rehash a user's password during verify if the hash cost is too low. [IDP-126](https://itse.youtrack.cloud/issue/IDP-126)

---

### Release PR Checklist
- [x] Update version number in main.php Sentry configuration


.github/pull_request_template.md
```diff
@@ -22,6 +22,7 @@
 - [ ] Documentation (README, local.env.dist, etc.)
 - [ ] Unit tests created or updated
 - [ ] Run `make composershow`
+- [ ] Run `make psr2`
```
jason-jackson, line 25: Probably doesn't matter right now as only 2 commands, but could make an additional make command to run all these rather than having to do `make composershow psr2 ...etc`

jason-jackson (approved):

application/console/controllers/GaController.php
```diff
@@ -43,7 +43,8 @@ public function actionRegister_event()
 
         $gaId = \Yii::$app->params['googleAnalytics']['measurementId'];
 
-        print_r(PHP_EOL .
+        print_r(
```
mtompset, line 46: print_r is ugly, but since it was already there... Otherwise, I don't see why an echo or print isn't sufficient in this case.

mtompset (approved):
Aim for shorter releases. :)

.github/pull_request_template.md
briskt: Will consider it for the future

application/console/controllers/GaController.php
briskt: Good to know.

briskt:
> Aim for shorter releases. :)

Will try, but I also don't want to slow down progress if reviews aren't forthcoming.

PR #327, opened 2024-02-21
convert from Codeship to GitHub Actions

briskt:
### Changed
- Use GitHub Actions in place of Codeship


actions-services.yml
```diff
@@ -0,0 +1,99 @@
...
+            EMAIL_SERVICE_baseUrl: http://email
+            EMAIL_SERVICE_validIpRanges: 192.168.0.0/16
+            EMAIL_SIGNATURE: Dummy Signature for Automated Tests
+            EMAILER_CLASS: \Sil\SilIdBroker\Behat\Context\fakes\FakeEmailer
+            HELP_CENTER_URL: https://help-center
+            IDP_NAME: Test
+            MYSQL_HOST: testdb
+            MYSQL_DATABASE: test
+            MYSQL_USER: idbroker
+            MYSQL_PASSWORD: idbroker
```
mtompset, line 31: even if it is test, shouldn't these sorts of things be hidden somehow?

mtompset: Oh, it was even in the codeship file. Isn't that still bad?

actions-services.yml
```diff
@@ -0,0 +1,99 @@
...
+            AWS_DEFAULT_REGION: us-east-1
+        command: "-jar DynamoDBLocal.jar -sharedDb"
+
+    # Don't use underscores in service names
+    mfaapi:
+        build: ./serverless-mfa-api
+        ports:
+            - 8080
+        environment:
+            AWS_ENDPOINT: dynamo:8000
```
jason-jackson, line 63: I think it would be worth using [extends](https://docs.docker.com/compose/multiple-compose-files/extends/#extending-services-within-the-same-file) and have a base service that has all the common env variables?

Something like
```
services:
  commonenv:
    environment:
      AWS_ACCESS_KEY_ID: abc123
      AWS_SECRET_ACCESS_KEY: abc123
      ...

  mfaapi:
    build: ./serverless-mfa-api
    extends: commonenv
    environment:
      NON_COMMON_VAR: foo
```

OR update file version to 3 and use [extension fields](https://docs.docker.com/compose/compose-file/11-extension/) (like auth-proxy).

jason-jackson (approved):

actions-services.yml
briskt: Yes, but the idea is to get the conversion done quickly and not go for optimization. Thanks for the suggestion.

actions-services.yml
briskt: What about this data gives you concern?

PR #329, opened 2024-02-27
use AWS AppConfig to load environment variables

briskt:
### Changed
- Read environment variables from AWS AppConfig at startup if APP_ID is defined

### Removed
- Removed the signal trap from the run script. It was broken a long while back by moving the apachectl command to the foreground, which was required to send logs to CloudWatch.

---

### Feature PR Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,5 +22,8 @@ COPY dockerbuild/vhost.conf /etc/apache2/sites-enabled/
 # ErrorLog inside a VirtualHost block is ineffective for unknown reasons
 RUN sed -i -E 's@ErrorLog .*@ErrorLog /proc/self/fd/2@i' /etc/apache2/apache2.conf
 
+ADD https://github.com/silinternational/config-shim/releases/latest/download/config-shim_0.0.2-pre.3_linux_amd64.tar.gz config-shim.tar.gz
+RUN tar xzf config-shim.tar.gz --directory /usr/local/bin --wildcards config-shim && rm config-shim*tar.gz
+
 EXPOSE 80
 CMD ["/data/run.sh"]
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -22,5 +24,8 @@ COPY dockerbuild/vhost.conf /etc/apache2/sites-enabled/
 # ErrorLog inside a VirtualHost block is ineffective for unknown reasons
 RUN sed -i -E 's@ErrorLog .*@ErrorLog /proc/self/fd/2@i' /etc/apache2/apache2.conf
 
+ADD https://github.com/silinternational/config-shim/releases/latest/download/config-shim_0.0.2-pre.3_linux_amd64.tar.gz config-shim.tar.gz
```
forevermatt, line 27: I see this URL uses both "latest" and "0.0.2-pre". As new versions are released, do you know if that will cause a problem with this URL?

application/run.sh
```diff
@@ -21,20 +18,22 @@ chown -R www-data:www-data \
...
+  if [[ ! -z $RUN_TASK ]]; then
     ./yii $RUN_TASK
     exit $?
-fi
+  fi
 
-apache2ctl -k start -D FOREGROUND
+  apache2ctl -k start -D FOREGROUND
+else
+  config-shim -v --app $APP_ID --config $CONFIG_ID --env $ENV_ID /data/yii migrate --interactive=0
```
forevermatt, line 31: Is there a way to have config-shim add the data to the existing environment, so that we don't have to prefix every command with the call to config-shim?

forevermatt (approved):
Seems like progress! (Had a couple of questions.)

Dockerfile
briskt: Good question. According to [this](https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases) I don't believe it will be a problem. I may also change the release script to remove the version number from the `tar.gz` file just to make it simpler.

application/run.sh
briskt: Not directly. It's not possible to change the environment of a parent process (e.g. bash) but it might be possible to write it to a file and then read from that file on each script line that needs it. I chose not to go down that road because it's more complex and because I would like to simplify this script in other ways such that it would only require one call to `config-shim`. What do you think?

forevermatt: Good call 👍

Dockerfile
forevermatt: Okay, thanks. Just wanted to ask.

PR #331, opened 2024-04-05
Release 6.7.1 -- better error, and auto-generate release string

briskt:
### Added
- Get the release tag from GitHub for use in the Sentry release string

### Fixed
- Fail earlier and produce a more appropriate error message if `employee_id` is not found.
- README correction


jason-jackson (approved):
Approved, but I'll have another PR probably later today if you want to hold off on the release to include that as well.

briskt:
> Approved, but I'll have another PR probably later today if you want to hold off on the release to include that as well.

I'm ok with smaller releases. Knowing another is coming, I won't deploy this to the masses of IdP instances, though.

PR #343, opened 2024-05-15
If an `authInit()` calls fails, skip loading that MFA's extra data

forevermatt:
### Changed (non-breaking)
- If an `authInit()` calls fails, skip loading that MFA's extra data
  * This should help us still allow a user to log in even if the WebAuthn MFA API is down, merely preventing those WebAuthn MFA options from working but allowing the "remember me" cookie and the other MFA options to work.

---

### Feature PR Checklist
- [x] Run `make composershow`
- [x] Run `make psr2`

### DIFF
--- a/application/common/models/Mfa.php
+++ b/application/common/models/Mfa.php
@@ -76,7 +76,16 @@ class Mfa extends MfaBase
     {
         $this->data = [];
         if ($this->verified === 1 && $this->scenario === User::SCENARIO_AUTHENTICATE) {
-            $this->data += $this->authInit($rpOrigin);
+            try {
+                $this->data += $this->authInit($rpOrigin);
+            } catch (\Exception $exception) {
+                \Yii::error([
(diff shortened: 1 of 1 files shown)


application/common/models/Mfa.php
```diff
@@ -76,7 +76,16 @@ public function loadData(string $rpOrigin = '')
...
         $this->data = [];
         if ($this->verified === 1 && $this->scenario === User::SCENARIO_AUTHENTICATE) {
-            $this->data += $this->authInit($rpOrigin);
+            try {
+                $this->data += $this->authInit($rpOrigin);
+            } catch (\Exception $exception) {
+                \Yii::error([
+                    'action' => 'load ' . $this->type . ' MFA data',
+                    'status' => 'error',
+                    'error' => 'authInit call failed (so skipping it): ' . $exception->getMessage(),
```
briskt, line 85: Very minor comment, but perhaps "authInit failed, bypassing this MFA option" might be a tad better.

PR #344, opened 2024-05-15
Release 6.7.5 - Enable login even if WebAuthn MFA API is down

forevermatt:
### Changed (non-breaking)
- If an `authInit()` calls fails, skip loading that MFA's extra data
  * This allows users to log in even if the WebAuthn MFA API is down, merely preventing those WebAuthn MFA options from working but allowing the "remember me" cookie and the other MFA options to work.

### Fixed
- Fix checklist entry and comment-typo


jason-jackson (approved):

application/common/models/Mfa.php
```diff
@@ -76,7 +76,16 @@ public function loadData(string $rpOrigin = '')
...
         $this->data = [];
         if ($this->verified === 1 && $this->scenario === User::SCENARIO_AUTHENTICATE) {
-            $this->data += $this->authInit($rpOrigin);
+            try {
+                $this->data += $this->authInit($rpOrigin);
+            } catch (\Exception $exception) {
+                \Yii::error([
+                    'action' => 'load ' . $this->type . ' MFA data',
+                    'status' => 'error',
+                    'error' => 'authInit call failed (so skipping it): ' . $exception->getMessage(),
```
briskt, line 85:
```suggestion
                    'error' => 'authInit failed, bypassing this MFA option: ' . $exception->getMessage(),
```
Very minor suggestion, feel free to proceed as-is.

briskt (approved):

application/common/models/Mfa.php
forevermatt: Hmm, thanks. I appreciate wording improvements. One tricky bit is that we are still including this MFA option in the list sent to our ssp container (so that the list of MFA IDs will matched what was used for the remember-me cookie), so I want to avoid implying that this MFA option was left out entirely. It will be present, but non-functional. I'll try a revised wording... please let me know what you think.

forevermatt: See #345 for suggested revision. Thoughts?

PR #351, opened 2024-08-14
Update base models, and do so with each "make composerupdate"

forevermatt:
### Added
- Revise Makefile to update base models after every composer update 
  * Occasionally Yii updates their template for Active Record model files. It seems more helpful to see the effects of those changes in that same PR (where Yii was updated), rather than seeing them in a later, unrelated PR. Updating the base models only takes around 10 seconds, so it seems like a minimal time cost (percentage-wise) that we're adding to each composer update call.

### Fixed
- Update the base models (using our current version of Yii). Insignificant changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ composershow:
 composerupdate:
 	docker compose run --rm cli composer update
 	make composershow
+	make basemodels
 
 db:
 	docker compose up -d db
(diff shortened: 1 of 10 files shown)


briskt:
>   * Occasionally Yii updates their template for Active Record model files. It seems more helpful to see the effects of those changes in that same PR (where Yii was updated), rather than seeing them in a later, unrelated PR. Updating the base models only takes around 10 seconds, so it seems like a minimal time cost (percentage-wise) that we're adding to each composer update call.

This reminds me that you have significant automation in mail-admin. Would any of that be of value in IdP repos? The one I'm most curious about is the one that updates dependencies on every release.

briskt (approved):

PR #352, opened 2024-08-19
[IDP-1176] Add `groups_external` field to User

forevermatt:
[IDP-1176](https://itse.youtrack.cloud/issue/IDP-1176) Add groups_external field to ID Broker's database

---

### Added
- Add `groups_external` field to User
- Add phpMyAdmin container for testdb

### Changed (non-breaking)
- Include `groups_external` values in a User's `member` list during login

---

### Feature PR Checklist
- [x] Unit tests created or updated
- [x] Run `make composershow`
- [x] Run `make psr2`

---

**Note:**
This does not yet provide any mechanism for _setting_ the new `groups_external` field. That will come in a subsequent PR.

### DIFF
--- a/application/common/models/User.php
+++ b/application/common/models/User.php
@@ -977,6 +977,7 @@ class User extends UserBase
         $labels['last_synced_utc'] = Yii::t('app', 'Last Synced (UTC)');
         $labels['created_utc'] = Yii::t('app', 'Created (UTC)');
         $labels['deactivated_utc'] = Yii::t('app', 'Deactivated (UTC)');
+        $labels['groups_external'] = Yii::t('app', 'Groups (External)');
 
         return $labels;
     }
--- a/application/common/models/User.php
+++ b/application/common/models/User.php
@@ -571,7 +571,15 @@ class User extends UserBase
             'member' => function (self $model) {
                 if (!empty($model->groups)) {
                     $member = explode(',', $model->groups);
+                } else {
+                    $member = [];
                 }
+
+                $externalGroups = explode(',', $model->groups_external);
+                foreach ($externalGroups as $externalGroup) {
+                    $member[] = $externalGroup;
+                }
+
                 $member[] = \Yii::$app->params['idpName'];
                 return $member;
             },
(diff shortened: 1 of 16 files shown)


forevermatt:
Fixing the bug (revealed by the tests) currently...

forevermatt:
Fixed.

application/console/migrations/m240813_155757_add_groups_external_field_to_user.php
```diff
@@ -0,0 +1,23 @@
...
+ * Class m240813_155757_add_groups_external_field_to_user
+ */
+class m240813_155757_add_groups_external_field_to_user extends Migration
+{
+    public function safeUp()
+    {
+        $this->addColumn(
+            '{{user}}',
+            'groups_external',
+            $this->string()->notNull()->defaultValue('')->after('groups')
```
mtompset, line 15: The Yii syntax is ugly.

mtompset (approved):
The difference between internal and external groups is unclear to me, but I wouldn't be surprised if it is somehow related to a Google API somewhere.

forevermatt:
> The difference between internal and external groups is unclear to me, but I wouldn't be surprised if it is somehow related to a Google API somewhere.

Thanks for mentioning that. The normal list of `groups` comes from the HR system, via ID Sync. This new `groups_external` field is for additional groups that need to be added from some other data source (e.g. a Google Sheet) that is external to the HR system.

application/console/migrations/m240813_155757_add_groups_external_field_to_user.php
briskt: PHP syntax is ugly. 😛

briskt (approved):

application/common/models/User.php
```diff
@@ -571,7 +571,17 @@ public function fields(): array
             'member' => function (self $model) {
                 if (!empty($model->groups)) {
                     $member = explode(',', $model->groups);
+                } else {
+                    $member = [];
                 }
```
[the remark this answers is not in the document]
forevermatt: Using a variable without defining it (e.g. assigning something to an array on an uninitialized array, in PHP) seems like a code smell. I simply added the `else` to ensure that the array would exist and be in the desired state, regardless of which path through the logic was executed.

Initializing the array first, as you suggested, is another valid option, but since it would be replaced with a different array within the if-block, I decided to only set it to an empty array if needed, rather than every time.

PR #355, opened 2024-08-20
[IDP-1182] Add endpoint for updating `groups_external`

forevermatt:
[IDP-1182](https://itse.youtrack.cloud/issue/IDP-1182) Provide a way to set `user.groups_external`

---

### Added
- Add an API endpoint (with documentation and tests) for updating a prefix-specific subset of a User's list of external groups

---

### Feature PR Checklist
- [x] Run `make composershow`
- [x] Run `make psr2`

### DIFF
--- a/api.raml
+++ b/api.raml
@@ -59,7 +59,7 @@ types:
     example: |
       {
         "email_address": "john_smith@example.com",
-        "app_prefix": "myapp"
+        "app_prefix": "myapp",
         "groups": ["myapp-users", "myapp-managers"]
       }
   UserResponse:
--- a/api.raml
+++ b/api.raml
@@ -44,6 +44,21 @@ types:
         "code": 0,
         "status": 400
       }
+  GroupsExternalUpdate:
+    type: object
+    properties:
+      email_address: string
+      app_prefix: string
+      groups:
+        type: string[]
+        description: >
+          The desired list of groups, including the app-prefix.
+    example: |
+      {
+        "email_address": "john_smith@example.com",
+        "app_prefix": "myapp"
+        "groups": ["myapp-users", "myapp-managers"]
+      }
   UserResponse:
     description: >
       Information on user record. Password is not included if a password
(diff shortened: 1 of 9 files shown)


api.raml
```diff
@@ -44,6 +44,24 @@ types:
...
+          The app-specific prefix, without the trailing hyphen.
+      groups:
+        type: string[]
+        description: >
+          The desired list of groups, including the app-prefix.
+    example: |
+      {
+        "email_address": "john_smith@example.com",
+        "app_prefix": "myapp",
+        "groups": ["myapp-users", "myapp-managers"]
```
briskt, line 63: Why is the prefix included as a separate property as well as included in the group names themselves?

application/common/models/User.php
```diff
@@ -1003,6 +1003,80 @@ public function isPromptForMfa(): bool
...
+                $this->addErrors([
+                    'groups_external' => sprintf(
+                        'The given group %s does not start with the given prefix (%s)',
+                        json_encode($appExternalGroup),
+                        json_encode($appPrefix)
+                    ),
+                ]);
+                return false;
+            }
+        }
```
briskt, line 1019: If the incoming group names omitted the prefix or the separate prefix property was omitted, this validation wouldn't be necessary. Doing so would remove one possible failure scenario.

briskt (approved):
Nicely organized implementation. My only question: what will be calling this endpoint?

forevermatt:
> Nicely organized implementation. My only question: what will be calling this endpoint?

I'll be writing a new custom sync process (unless one we already have will suffice) for syncing from a Google Sheet to this new field, via this new API endpoint.

api.raml
forevermatt: The new sync process (yet to be written) will have a single, set prefix (per instance), so that a given copy of the new sync process can only affect external groups with that prefix.

The current need is for syncing from a Google Sheet controlled by some of the REAP admins, to enable them to add "reap-____" groups to users signing into REAP (because the new platform, Preservica, requires the use of SAML-provided groups for controlling authorization). This will let them add specific groups to specific users in specific IDPs, to control what those users are allowed to do in REAP. The `app_prefix` will be controlled by us, who set up and run the instance of the new sync process. The `groups` are controlled by the REAP admins (via the Google Sheet), so this is just another way to ensure the data they provide meets the constraints we have established for this new part of our system.

I hope that made sense. Feel free to ask more questions and/or review the parent YouTrack issue ([IDP-1153](https://itse.youtrack.cloud/issue/IDP-1153)) for more details.

application/common/models/User.php
forevermatt: True. I considered that option, but I'd like the group names (as typed into the Google Sheet) to _exactly_ match the group names that REAP will receive via SAML, to help make it more intuitive.

If I had them simply type in everything after the prefix (on the Google Sheet), then they would type in something like "europe" and have to know/remember that it will come to REAP via SAML as "reap-europe". It seemed better to keep that prefix as visible as possible, so that they don't set up rules for a "europe" group and then be confused when it doesn't work.

PR #358, opened 2024-08-28
[IDP-1183, partial] Revise "external groups" API endpoints

forevermatt:
[IDP-1183](https://itse.youtrack.cloud/issue/IDP-1183) Sync prefixed-groups from Google Sheet to ID Broker's `groups_external` field

---

### Added
- Add a way to run the personnel-sync locally, for dev and testing
- Add an API endpoint for listing users with groups with a specific prefix
- Add separate API endpoint for creating a user's prefixed external groups 
  * This avoids the need to include the email address in the URL path for creates, which personnel-sync does not currently support.
  * This actually makes the `POST` behavior more typical, as you might normally POST to an endpoint for a type of resource (without a unique identifier), rather than posting to a specific resource.
- Add tests for these various API endpoints

### Changed (non-breaking)
- Put email and app-prefix in URL for update-external-groups API
  * This _would have been_ a breaking change, but we had not yet released that new API endpoint.
- Use comma-delimited string for external groups in API endpoints
  * This also _would have been_ a breaking change, if the previous version had already been released.
  * This makes it easier to sync this data with a spreadsheet (e.g. a Google Sheet), where values are just strings, not arrays.
- Enable updating a user's external groups for an app to be an empty list

---

### Feature PR Checklist
- [x] Run `make composershow`
- [x] Run `make psr2`


application/common/models/User.php
```diff
@@ -641,6 +641,28 @@ public function getNagState()
         return $this->nagState->getState();
     }
 
+    private static function listUsersWithExternalGroupWith($appPrefix): array
+    {
+        $appPrefixWithHyphen = $appPrefix . '-';
+
+        /** @var User[] $users */
+        $users = User::find()->where(
+            ['like', 'groups_external', $appPrefixWithHyphen]
```
briskt, line 650: Doesn't this need a `%` before and after `$appPrefixWithHyphen`? Otherwise it's only going to find records where `groups_external` is exactly 'groupname-' I think. _...some time later..._ I did some reading in the Yii docs and some other code, and I am still puzzled. It appears that Yii magically adds the %, but it is not exactly documented as such. All that to say is that I think this code is correct, but I don't know why.

application/common/models/User.php
```diff
@@ -1003,8 +1025,46 @@ public function isPromptForMfa(): bool
         return false;
     }
 
-    public function updateExternalGroups($appPrefix, $appExternalGroups): bool
+    public static function syncExternalGroups(string $appPrefix, array $desiredExternalGroupsByUserEmail): array
```
briskt, line 1028: This could benefit from a function doc header since "sync external groups" sounds like it's going to sync data to groups that are external, i.e. outside of idp-id-broker. In contrast, it's actually pulling **in** data and updating user data. Here's a start:
```php
/*
 * syncExternalGroups updates users' `externalGroups` attribute with data in the
 * `desiredExternalGroupsByUserEmail` parameter. That parameter is indexed by user 
 * email address and contains a list of external groups to add to the `externalGroups`
 * attribute.
 */
```
It may also be helpful to rename the function to something like `updateUsersExternalGroups`.

application/common/models/User.php
```diff
@@ -1003,8 +1025,46 @@ public function isPromptForMfa(): bool
...
+            }
+        }
+
+        foreach ($desiredExternalGroupsByUserEmail as $email => $groupsForPrefix) {
+            $user = User::findByEmail($email);
+            if ($user === null) {
+                $errors[] = 'No user found for email address ' . json_encode($email);
+                continue;
+            }
+            $successful = $user->updateExternalGroups($appPrefix, $groupsForPrefix);
```
briskt, line 1047: I could be wrong, but the `app` part of the `appPrefix` parameter name seems out of place. It may have meaning elsewhere, but it just raises questions in my mind. My 2 cents.

briskt (approved):
Looks good. I offer a couple suggestions for documentation and naming. Take 'em or leave 'em.

application/common/models/User.php
```diff
@@ -641,6 +641,28 @@ public function getNagState()
         return $this->nagState->getState();
     }
 
+    private static function listUsersWithExternalGroupWith($appPrefix): array
```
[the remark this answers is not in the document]
forevermatt: In general, yes: I'm trying to place new functions alphabetically. It's a bit subjective since many of the other functions aren't alphabetical, but it seems to reduce conflicts that would have occurred from multiple people just putting new functions at the end. 🤷‍♂️ 

And regarding the name, my thinking is this: A function named `listSomething()` is getting a list of identifiers of some sort, rather than getting the full objects themselves (vs. `getTheThings()`, for example). I can go with the more specific name (`listUserEmailsWith...`), as that does improve clarity, which is certainly valuable. I'm just trying to balance that against super-long names.

If you have a strong preference, please feel free to say so. For now, I think I'll leave it as-is, though I'm open to changing it.

application/common/models/User.php
forevermatt: Yii automatically wraps the `like` comparison's value in `%`s. Here are the docs (though I agree they're not always easy to find):
https://www.yiiframework.com/doc/guide/2.0/en/db-query-builder#operator-format

application/common/models/User.php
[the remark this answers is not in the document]
briskt: I guess the type system clarifies what is actually returned, so leaving it is fine.

application/common/models/User.php
forevermatt: Ahh, good point. I started adding docs in a subsequent branch (not yet PR'd), but I appreciate the insights on helpful wording/naming. 👍

application/common/models/User.php
forevermatt: Yeah, I've been having trouble deciding on a good name for that. It's the group-name prefix for a particular application (e.g. "wiki").

This `groups_external` field can hold groups used by multiple different applications (SPs). For example, a User's `groups_external` field might have `wiki-managers,map-america,map-europe` (which would be a "wiki-managers" group for use by the "wiki" application, and "map-america" and "map-europe" to be used by the "map" application, or at least that's the intent behind this design.

If given enough context, I can simply call it the prefix... and perhaps I should just use that everywhere.

Thanks for sharing your 2 cents. That's often helpful.

PR #360, opened 2024-09-05
[IDP-1183, partial] Add cron-based sync of `groups_external` data from Google Sheet

forevermatt:
[IDP-1183](https://itse.youtrack.cloud/issue/IDP-1183) Sync prefixed-groups from Google Sheet to ID Broker's `groups_external` field

---

### Added
- Add cron-based sync of `groups_external` data from Google Sheet

### Fixed
- Clean up some external-groups log message formatting
- Remove redundant log message from `$user->updateExternalGroups()`

---

### Feature PR Checklist
- [x] Run `make composershow`
- [x] Run `make psr2`

### DIFF
--- a/application/common/components/ExternalGroupsSync.php
+++ b/application/common/components/ExternalGroupsSync.php
@@ -3,6 +3,7 @@
 namespace common\components;
 
 use common\models\User;
+use Webmozart\Assert\Assert;
 use Yii;
 use yii\base\Component;
 
@@ -49,7 +50,7 @@ class ExternalGroupsSync extends Component
     private static function syncSet(string $appPrefix, string $googleSheetId)
     {
         $desiredExternalGroups = self::getExternalGroupsFromGoogleSheet($googleSheetId);
-        $errors = User::syncExternalGroups($appPrefix, $desiredExternalGroups);
+        $errors = User::updateUsersExternalGroups($appPrefix, $desiredExternalGroups);
         Yii::warning(sprintf(
             "Ran sync for '%s' external groups.",
             $appPrefix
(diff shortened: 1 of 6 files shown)


application/common/components/ExternalGroupsSync.php
```diff
@@ -0,0 +1,110 @@
...
+        Yii::warning(sprintf(
+            "Ran sync for '%s' external groups.",
+            $appPrefix
+        ));
+
+        if (! empty($errors)) {
+            Yii::error(sprintf(
+                'Errors that occurred while syncing %s external groups: %s',
+                $appPrefix,
+                join(" / ", $errors)
```
briskt, line 63: Since it's theoretically possible to accumulate thousands of errors, would it be good to limit the length of this log message?

briskt (approved):

application/common/components/ExternalGroupsSync.php
```diff
@@ -0,0 +1,110 @@
...
+            $appPrefixKey = sprintf('set%uAppPrefix', $i);
+            $googleSheetIdKey = sprintf('set%uGoogleSheetId', $i);
+
+            if (! array_key_exists($appPrefixKey, $syncSetsParams)) {
+                Yii::warning(sprintf(
+                    'Finished syncing external groups after %s sync set(s).',
+                    ($i - 1)
+                ));
+                break;
+            }
```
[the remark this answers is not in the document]
forevermatt: That makes sense, but since this is also used for the names of the environment variables, starting with 1 seems more intuitive than starting with 0 in this case. I agree that it's more intuitive/expected that code-only for-loops start with 0, though.

application/common/components/ExternalGroupsSync.php
forevermatt: Hmm, I hadn't thought of that. Yes, I can limit that error message length.

forevermatt: #361

PR #366, opened 2024-09-11
Idp 781 update last login after mfa

hobbitronics:
[IDP-781](https://itse.youtrack.cloud/issue/IDP-781)

### Fixed
- update last_login_utc after mfa verification

---

### Feature PR Checklist
- [ ] Run `make composershow`
- [ ] Run `make psr2`

### DIFF
--- a/application/frontend/controllers/MfaController.php
+++ b/application/frontend/controllers/MfaController.php
@@ -5,7 +5,7 @@ namespace frontend\controllers;
 use common\models\Mfa;
 use common\models\MfaWebauthn;
 use common\models\User;
-use common\components\MfaBackendWebAuthn;
+use common\helpers\MySqlDateTime;
 use frontend\components\BaseRestController;
 use stdClass;
 use yii\web\BadRequestHttpException;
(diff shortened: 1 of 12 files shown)


application/frontend/controllers/MfaController.php
```diff
@@ -115,6 +115,14 @@ public function actionVerify(int $id, string $type = "")
             throw new BadRequestHttpException();
         }
         $mfa->loadData($rpOrigin);
+        $user->last_login_utc = MySqlDateTime::now();
```
briskt, line 118: I think this controller action is also used to verify a new webauthn, but I could be wrong.

briskt (commented):
It's going to be more tricky than this, because the same endpoint is used to verify a new webauthn, and we probably don't want to set the date in that case. Though I suppose it wouldn't be terrible since they would have had to login not long prior to that time in order to add the webauthn.

application/frontend/controllers/MfaController.php
hobbitronics: It looks that way from the tests. Any ideas on how to only update it when logging in? I was wondering about only updating if the previous one is older than the token lifespan.

hobbitronics:
> It's going to be more tricky than this, because the same endpoint is used to verify a new webauthn, and we probably don't want to set the date in that case. Though I suppose it wouldn't be terrible since they would have had to login not long prior to that time in order to add the webauthn.

We would also need to update the last_login_utc when mfa isn't required (30 days or so after using it). That data doesn't seem to be available here unless its stored in the user. I am guessing it is stored as a cookie or in a session by simpleSaml on the client. I am starting to think this is the tricky part.

PR #375, opened 2024-09-30
[IDP-1221] Optionally send external-groups sync-errors notification email

forevermatt:
[IDP-1221](https://itse.youtrack.cloud/issue/IDP-1221) Send error notification email for external-groups sync errors

---

### Added
- Optionally send external-groups sync-errors notification email (see `local.env.dist` for the environment variable name)

### Changed (non-breaking)
- Enable the `Emailer` to send to an arbitrary email address

### Fixed
- Update dependencies

---

### Feature PR Checklist
- [x] Run `make composershow`
- [x] Run `make psr2`

### DIFF
--- a/application/common/mail/ext-group-sync-errors.html.php
+++ b/application/common/mail/ext-group-sync-errors.html.php
@@ -37,4 +37,8 @@ if (empty ($googleSheetUrl)) {
 }
 ?>
 
+<p>
+    Other users' external groups may have been synced successfully.
+</p>
+
 <p><i><?= nl2br(yHtml::encode($emailSignature), false) ?></i></p>
--- a/application/common/mail/ext-group-sync-errors.html.php
+++ b/application/common/mail/ext-group-sync-errors.html.php
@@ -17,7 +17,7 @@ use yii\helpers\Html as yHtml;
 <?= yHtml::ul($errors) ?>
 
 <?php
-if (empty ($googleSheetUrl)) {
+if (empty($googleSheetUrl)) {
     ?>
     <p>
         If any of these seem like simple data problems, you can potentially fix them
--- a/application/common/mail/ext-group-sync-errors.html.php
+++ b/application/common/mail/ext-group-sync-errors.html.php
@@ -1,0 +1,40 @@
+<?php
+use yii\helpers\Html as yHtml;
+
+/**
+ * @var string $appPrefix
+ * @var string[] $errors
+ * @var string $googleSheetUrl
+ * @var string $emailSignature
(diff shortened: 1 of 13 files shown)


briskt:
I'm not opposed to a breaking change, but would like to consider other things to include in the next major version. If this is a time-sensitive change, it might be good to do it in a non-breaking way and then circle back with the improved code.

forevermatt:
> I'm not opposed to a breaking change, but would like to consider other things to include in the next major version. If this is a time-sensitive change, it might be good to do it in a non-breaking way and then circle back with the improved code.

Good point. Changing the authenticate call to no longer update the user's last-login date/time will probably also be a breaking change, so it may be best to hold off on this minimum-version update until that is ready to release (and then perhaps raise our minimum version all the way to the oldest still-supported version of PHP).

I'll look at adjusting this code back down to be compatible with PHP 7.4

forevermatt:
> I'm not opposed to a breaking change, but would like to consider other things to include in the next major version. If this is a time-sensitive change, it might be good to do it in a non-breaking way and then circle back with the improved code.

Okay, I've lowered the minimum PHP version for ID Broker back to 7.4, to avoid the breaking change in this PR.

briskt:
> Changing the authenticate call to no longer update the user's last-login date/time will probably also be a breaking change

My perspective would be that this is a bug fix, so it wouldn't be a breaking change. But it's a bit subjective.

application/common/mail/ext-group-sync-errors.html.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+?>
+<p>
+    The following errors occurred when syncing the <?= yHtml::encode($appPrefix) ?>
+    external groups to the <?= yHtml::encode($idpDisplayName) ?> IDP:
+</p>
+<?= yHtml::ul($errors) ?>
+
+<?php
+if (empty($googleSheetUrl)) {
```
briskt, line 20: Isn't it impossible for `googleSheetUrl` to be empty? It would at least have this much: 'https://docs.google.com/spreadsheets/d/'

briskt (approved):

application/common/mail/ext-group-sync-errors.html.php
forevermatt: Ahh, good catch. I'll try to fix that.

forevermatt: Submitted follow-up PR to fix that:
https://github.com/silinternational/idp-id-broker/pull/376

PR #379, opened 2024-10-21
Pushing Docker Image to GHCR.io

Praveenraj-K:
### Added
- Added step to push the docker image to Github Container Registry


briskt (approved):
Looks good. Also remember that repos with a develop branch should be merged from a feature branch. I understand it's confusing because we have these two different types of repos.

PR #382, opened 2024-10-29
Release 6.10.4 - Update dependencies

forevermatt:
### Fixed
- Update dependencies

---

### PR Checklist
- [x] Run `make composershow`
- [x] Run `make psr2`

### DIFF
--- a/application/composer.lock
+++ b/application/composer.lock
@@ -676,16 +676,16 @@
         },
         {
             "name": "google/apiclient-services",
-            "version": "v0.375.0",
+            "version": "v0.379.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/googleapis/google-api-php-client-services.git",
-                "reference": "ad8c4fcf925fd10198c20e2d7a3390183dc24bf1"
+                "reference": "5a671ff040d55cec7b69b88d84c91bd8867d1c8e"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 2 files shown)


briskt (approved):
Looks like we need to fix the workflow to not push to GHCR on feature branches, or we need to not use slashes in branch names.

forevermatt:
> Looks like we need to fix the workflow to not push to GHCR on feature branches, or we need to not use slashes in branch names.

Yeah, I noticed that and was wondering what to do about it. Apparently there's an output you can use for getting those tags (hopefully in a format that works for tags, i.e. replacing slashes with hyphens). I'm also fine with only pushing docker images for the `main` branch, if you prefer that approach.

forevermatt:
> > Looks like we need to fix the workflow to not push to GHCR on feature branches, or we need to not use slashes in branch names.
> 
> Yeah, I noticed that and was wondering what to do about it. Apparently there's an output you can use for getting those tags (hopefully in a format that works for tags, i.e. replacing slashes with hyphens). I'm also fine with only pushing docker images for the `main` branch, if you prefer that approach.

Here's the approach using and output to get the tags, but I don't know if we need to upgrade to version 6 of that GitHub Action to be able to use it (I would _guess_ not...):
https://docs.docker.com/build/ci/github-actions/manage-tags-labels/#:~:text=steps.meta.outputs.tags

briskt:
> I'm also fine with only pushing docker images for the `main` branch, if you prefer that approach.

That's the direction I've been going lately.

PR #383, opened 2024-11-05
Release 6.10.5 - Avoid sending abandoned-users email to HR too frequently

forevermatt:
[IDP-1266](https://itse.youtrack.cloud/issue/IDP-1266) Run the external-groups sync every 12 hours

* This is pre-work, to enable us to avoid sending both the abandoned-users email and the external-groups sync-errors email too frequently.
* Another PR will follow that "rate limits" the external-groups sync-errors email.
* That, in turn, will allow us to run the cron as frequently as we want (e.g. every 12 hours).

---

### Changed (non-breaking)
- Switch `EmailLog` from tracking by `user_id` to by `to_address`
- Log all emails sent, not just those to Users in our database

### Fixed
- Add 'ext-group-sync-errors' and 'abandoned-users' EmailLog message types
- Fix argument within `iCheckIfAGetBackupCodesEmailHasBeenSentRecently()`
  * It was passing a `User` when it should have been passing the User ID.
- Use reflection to avoid need to maintain duplicate list of message types
- Avoid sending abandoned-users email to HR too frequently

---

### PR Checklist
- [x] Run `make composershow`
- [x] Run `make psr2`

### DIFF
--- a/application/common/components/Emailer.php
+++ b/application/common/components/Emailer.php
@@ -313,6 +313,7 @@ class Emailer extends Component
             EmailLog::MESSAGE_TYPE_MFA_RATE_LIMIT => $this->subjectForMfaRateLimit,
             EmailLog::MESSAGE_TYPE_PASSWORD_CHANGED => $this->subjectForPasswordChanged,
             EmailLog::MESSAGE_TYPE_WELCOME => $this->subjectForWelcome,
+            EmailLog::MESSAGE_TYPE_ABANDONED_USERS => $this->subjectForAbandonedUsers,
             EmailLog::MESSAGE_TYPE_EXT_GROUP_SYNC_ERRORS => $this->subjectForExtGroupSyncErrors,
             EmailLog::MESSAGE_TYPE_GET_BACKUP_CODES => $this->subjectForGetBackupCodes,
             EmailLog::MESSAGE_TYPE_REFRESH_BACKUP_CODES => $this->subjectForRefreshBackupCodes,
@@ -381,6 +382,13 @@ class Emailer extends Component
         $ccAddress = $data['ccAddress'] ?? '';
         $bccAddress = $data['bccAddress'] ?? '';
         $subject = $this->getSubjectForMessage($messageType, $dataForEmail);
+
+        if (empty($subject)) {
+            \Yii::error(sprintf(
+                'No subject known for %s email messages.',
+                $messageType
+            ));
+        }
 
         $this->email($toAddress, $subject, $htmlBody, strip_tags($htmlBody), $ccAddress, $bccAddress, $delaySeconds);
 
--- a/application/common/components/Emailer.php
+++ b/application/common/components/Emailer.php
@@ -385,7 +385,7 @@ class Emailer extends Component
         $this->email($toAddress, $subject, $htmlBody, strip_tags($htmlBody), $ccAddress, $bccAddress, $delaySeconds);
 
         if ($user !== null) {
-            EmailLog::logMessage($messageType, $user->id);
+            EmailLog::logMessage($messageType, $toAddress);
         }
     }
 
@@ -410,17 +410,39 @@ class Emailer extends Component
     }
 
     /**
-     *
      * Whether the user has already been sent this type of email in the last X days
      *
      * @param int $userId
      * @param string $messageType
      * @return bool
      */
-    public function hasReceivedMessageRecently($userId, string $messageType)
+    public function hasUserReceivedMessageRecently($userId, string $messageType): bool
     {
-        $latestEmail = EmailLog::find()->where(['user_id' => $userId, 'message_type' => $messageType])
-            ->orderBy('sent_utc DESC')->one();
+        $user = User::findOne($userId);
+        if (empty($user)) {
+            \Yii::warning(sprintf(
+                'User (%s) not found when checking for recent emails',
+                $userId
+            ));
+            return false;
+        }
+        return $this->hasAddressReceivedMessageRecently($user->email, $messageType);
+    }
+
+    public function hasAddressReceivedMessageRecently(?string $toAddress, string $messageType): bool
+    {
+        if (empty($toAddress)) {
+            \Yii::warning('Not given any email address when checking for recent emails');
+            return false;
+        }
+
+        $latestEmail = EmailLog::find()->where([
+            'to_address' => $toAddress,
+            'message_type' => $messageType,
+        ])->orderBy(
+            'sent_utc DESC'
+        )->one();
+
         if (empty($latestEmail)) {
             return false;
         }
@@ -486,7 +508,7 @@ class Emailer extends Component
         return $this->sendGetBackupCodesEmails
             && $user->getVerifiedMfaOptionsCount() === 1
             && !$user->hasMfaBackupCodes()
-            && !$this->hasReceivedMessageRecently($user->id, EmailLog::MESSAGE_TYPE_GET_BACKUP_CODES);
+            && !$this->hasUserReceivedMessageRecently($user->id, EmailLog::MESSAGE_TYPE_GET_BACKUP_CODES);
     }
 
     /**
@@ -513,7 +535,7 @@ class Emailer extends Component
             return false;
         }
 
-        if ($this->hasReceivedMessageRecently($user->id, EmailLog::MESSAGE_TYPE_LOST_SECURITY_KEY)) {
+        if ($this->hasUserReceivedMessageRecently($user->id, EmailLog::MESSAGE_TYPE_LOST_SECURITY_KEY)) {
             return false;
         }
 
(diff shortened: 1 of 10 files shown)


forevermatt:
Sorry, this is a medium-long PR.

And the failed build-and-publish is because of our bug in tag names we're trying to push to GHCR (for "feature/..." branches).

application/common/components/Emailer.php
```diff
@@ -410,24 +421,66 @@ public function sendDelayedMfaRelatedEmails()
...
-            ->orderBy('sent_utc DESC')->one();
+        $user = User::findOne($userId);
+        if (empty($user)) {
+            \Yii::warning(sprintf(
+                'User (%s) not found when checking for recent emails',
+                $userId
+            ));
+            return false;
+        }
+        return $this->hasAddressReceivedMessageRecently($user->email, $messageType);
```
briskt, line 440: Probably should use `getEmailAddress` rather than accessing the `email` property directly.
```suggestion
        return $this->hasAddressReceivedMessageRecently($user->getEmailAddress(), $messageType);
```

application/common/components/Emailer.php
```diff
@@ -410,24 +421,66 @@ public function sendDelayedMfaRelatedEmails()
...
+    }
+
+    public function hasAddressReceivedMessageRecently(?string $toAddress, string $messageType): bool
```
briskt, line 443: You can avoid the `?` on the parameter type. The email address is never null, I believe.
```suggestion
    public function hasAddressReceivedMessageRecently(string $toAddress, string $messageType): bool
```

application/common/components/Emailer.php
```diff
@@ -382,11 +388,16 @@ public function sendMessageTo(
         $bccAddress = $data['bccAddress'] ?? '';
         $subject = $this->getSubjectForMessage($messageType, $dataForEmail);
 
+        if (empty($subject)) {
+            \Yii::error(sprintf(
+                'No subject known for %s email messages.',
+                $messageType
+            ));
+        }
```
briskt, line 396: Can/should this check be done within `email`?

application/common/components/Emailer.php
```diff
@@ -410,24 +421,66 @@ public function sendDelayedMfaRelatedEmails()
...
+            ));
+            return false;
+        }
+        return $this->hasAddressReceivedMessageRecently($user->email, $messageType);
+    }
+
+    public function hasAddressReceivedMessageRecently(?string $toAddress, string $messageType): bool
+    {
+        if (empty($toAddress)) {
+            \Yii::warning('Not given any email address when checking for recent emails');
```
briskt, line 446: Not sure this is the best idea, but I'll suggest anyway. Remove this warning and then you can drop the duplicate check in `shouldSendAbandonedUsersMessage`.
```suggestion
```

application/common/components/Emailer.php
```diff
@@ -410,24 +421,66 @@ public function sendDelayedMfaRelatedEmails()
...
+     * Whether we should send an abandoned-users message to HR.
+     *
+     * @return bool
+     */
+    public function shouldSendAbandonedUsersMessage(): bool
+    {
+        if (empty($this->hrNotificationsEmail)) {
+            return false;
+        }
+
```
briskt, line 474: See earlier comment about removing the warning.
```suggestion
```

application/common/models/EmailLog.php
```diff
@@ -44,43 +49,40 @@ public function attributeLabels()
...
-            self::MESSAGE_TYPE_METHOD_PURGED,
-            self::MESSAGE_TYPE_MFA_MANAGER,
-            self::MESSAGE_TYPE_MFA_MANAGER_HELP,
-            self::MESSAGE_TYPE_PASSWORD_EXPIRING,
-            self::MESSAGE_TYPE_PASSWORD_EXPIRED,
-            self::MESSAGE_TYPE_PASSWORD_PWNED,
-        ];
+        $reflectionClass = new ReflectionClass(__CLASS__);
+        $messageTypes = [];
+        foreach ($reflectionClass->getConstants() as $name => $value) {
```
briskt, line 56: Nice!

application/common/models/User.php
```diff
@@ -1278,6 +1281,16 @@ public function getEmailAddress(): string
...
     }
 
+    /**
+     * Gets query for [[EmailLogs]].
+     *
+     * @return \yii\db\ActiveQuery
+     */
+    public function getEmailLogs()
+    {
+        return $this->hasMany(EmailLog::class, ['to_address' => 'email']);
```
briskt, line 1291: Will this work if you replace `'email'` with `$this->getEmailAddress()`?

application/common/models/User.php
```diff
@@ -20,6 +20,9 @@
 use yii\db\conditions\OrCondition;
 use yii\helpers\ArrayHelper;
 
+/**
+ * @property EmailLog[] $emailLogs
```
briskt, line 24: I _think_ the only place the `EmailLog` properity would be used is in the `beforeDelete` method. It may be simpler, for both implementation and maintenance, to just update that method with the appropriate query.

application/console/migrations/m241029_202830_refactor_email_log_to_use_address.php
```diff
@@ -0,0 +1,37 @@
...
+        $this->addColumn(
+            '{{email_log}}',
+            'to_address',
+            $this->string()->notNull()->defaultValue('')
+        );
+
+        // Populate the `to_address` column for existing records.
+        $this->execute('
+            UPDATE `email_log`, `user`
+            SET `email_log`.`to_address` = `user`.`email`
```
briskt, line 23: This will be correct for most of the records, but there may be some that were sent to the `personal_email`. Not likely, or maybe impossible, for it to be any of those that we're checking recently sent. But a complication to wrestle with.

briskt (commented):
It all looks very good in terms of code quality and creative solutions. But there's a complication in that users may not actually have a valid `email` property. This affects sending, but more importantly, it affects the email log. Let me know if you'd like to talk through options.

application/common/models/User.php
forevermatt: I'd like to keep using Yii's built-in capability for handling those model-to-model relations. I originally defined this as a foreign key, but realized that wouldn't work well (for various reasons), but that's where I got the code for how to set up the code so that `User` models can easily get the list of `emailLogs` that match that User's (current) email address.

application/common/components/Emailer.php
forevermatt: Good question. I think it belongs here, since this is where the `getSubjectForMessage()` call is happening. When `email()` is called, it's being handed a pre-rendered subject. This warning is about when we're trying to look up the subject for a given message type, but no such subject was found. (I hit that situation before realizing I needed to add the new subjects to the `subjects` array internally; in that case, the subject just comes back as an empty string, which was not helpful.)

I suppose I could move it into `getSubjectForMessage()`... that actually may make more sense than where I have it here. I'll probably do that, now that you mention the question of where this check would best be placed.

application/common/components/Emailer.php
forevermatt: Ahh, thanks. Good catch.

application/common/models/EmailLog.php
mtompset: Yes, flexible code for when some other MESSAGE_TYPE_ is added.

application/common/components/Emailer.php
forevermatt: I think (as you say) we are not ever currently passing in anything other than a string. However, depending on that seems a little fragile. For example, the `hrNotificationsEmail` value is initially null. We have a check in `Emailer->init()` to see if it is, and if so change it to an empty string, but there's nothing about the property that guarantees that it's always a string.

It seems more resilient to allow this to be null, and simply handle it internally (as the `empty` check does), if you don't mind me leaving it this way.

application/common/models/User.php
mtompset: I find it weird that this isn't in a BASE model, and not in an extended model class.

application/common/components/Emailer.php
forevermatt: I'm not sure I understand. This seems like a check that this function should be responsible for doing, to ensure it properly handles any unacceptable values it might be passed, rather than depending on all of the other code that might call this to do the data validation for it. I'd also like to avoid too tightly coupling this function's logic with that of the `shouldSendAbandonedUsersMessage()` function.

application/common/components/Emailer.php
briskt: That's fine, thanks.

application/common/components/Emailer.php
briskt: My comment was simply reducing complexity, with little to no consideration of coupling. Do what you think is best.

application/common/models/User.php
briskt: As we discussed on Meet, changing the migration will solve this.

application/console/migrations/m241029_202830_refactor_email_log_to_use_address.php
briskt: Yes, I think it's safe to ignore the email change corner case.

application/common/components/Emailer.php
mtompset: debugging parameter type issues is painful!

forevermatt:
Closing in favor of #384 (and subsequent PRs) instead

PR #387, opened 2024-11-12
[IDP-1266, part 2] Rate limit both non user emails, add EmailLog validation

forevermatt:
[IDP-1266](https://itse.youtrack.cloud/issue/IDP-1266) (partial:) Run the external-groups sync every 12 hours

---

### Changed (non-breaking)
- Avoid sending 'ext-groups-sync-errors' email too frequently

### Fixed
- Add a test that the 'abandoned-user' email isn't sent too frequently
- Add a test that the 'ext-groups-sync-errors' email isn't sent too frequently
- Remove unneeded use import in a test file
- Ensure we don't log an email as to both a User and non-user address
- Ensure every email we log is to either a User or a non-user address

---

### PR Checklist
- [ ] ~~Put version number in PR title (e.g. `Release x.y.z - Summary of changes`)~~
- [ ] ~~Documentation (README, local.env.dist, api.raml, etc.)~~
- [x] Run `make composershow`
- [x] Run `make psr2`

### DIFF
--- a/application/common/models/EmailLog.php
+++ b/application/common/models/EmailLog.php
@@ -122,6 +122,11 @@ class EmailLog extends EmailLogBase
                 'non_user_address',
                 'emptyIfUserIsSet',
             ],
+            [
+                'non_user_address',
+                'requiredIfUserIsEmpty',
+                'skipOnEmpty' => false,
+            ],
         ], parent::rules());
     }
 
@@ -132,6 +137,19 @@ class EmailLog extends EmailLogBase
                 $attribute,
                 sprintf(
                     'Only a user_id or a %s may be set, not both',
+                    $attribute
+                )
+            );
+        }
+    }
+
+    public function requiredIfUserIsEmpty($attribute)
+    {
+        if (empty($this->user_id) && empty($this[$attribute])) {
+            $this->addError(
+                $attribute,
+                sprintf(
+                    'Either a user_id or a %s must be set',
                     $attribute
                 )
             );
--- a/application/common/models/EmailLog.php
+++ b/application/common/models/EmailLog.php
@@ -118,6 +118,23 @@ class EmailLog extends EmailLogBase
                 'default',
                 'value' => MySqlDateTime::now(),
             ],
+            [
+                'non_user_address',
+                'emptyIfUserIsSet',
+            ],
         ], parent::rules());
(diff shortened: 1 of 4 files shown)


application/common/models/EmailLog.php
```diff
@@ -118,6 +118,41 @@ public function rules()
...
+            ],
+            [
+                'non_user_address',
+                'requiredIfUserIsEmpty',
+                'skipOnEmpty' => false,
+            ],
         ], parent::rules());
     }
+
+    public function emptyIfUserIsSet($attribute)
```
briskt, line 133: These validators don't read well as standalone functions. At least it took me more time to figure them out for as simple as they should be. I couldn't think of a better name for the function so I just wrote a function header.
```suggestion

    /**
     * Validate an attribute to ensure it is empty (unset) if `user_id` is set.
     *
     * @param string $attribute Attribute name to be validated
     * @return bool
     */
    public function emptyIfUserIsSet($attribute)
```

application/common/models/EmailLog.php
```diff
@@ -118,6 +118,41 @@ public function rules()
...
+                $attribute,
+                sprintf(
+                    'Only a user_id or a %s may be set, not both',
+                    $attribute
+                )
+            );
+        }
+    }
+
+    public function requiredIfUserIsEmpty($attribute)
```
briskt, line 146:
```suggestion

    /**
     * Validate an attribute to ensure it is not empty if `user_id` is not set.
     *
     * @param string $attribute Attribute name to be validated
     * @return bool
     */
    public function requiredIfUserIsEmpty($attribute)
```

briskt (approved):
Nice work

application/common/models/EmailLog.php
mtompset: Actually, using a validator class for each function is a better implementation, in my opinion. Individual functions like this are so Yii1.

forevermatt: Thanks, @briskt.

And you're right, @mtompset: a validator class is probably a better way to do this. I opted for the in-class validator function for these, since it was a smaller step to take. We'll try proceeding with the DocBlock that @briskt added, but if these getting any bigger, moving them to validator classes is a good idea.

PR #392, opened 2024-12-11
Fix Sentry logging details

briskt:
[IDP-779](https://itse.youtrack.cloud/issue/IDP-779) Sentry log events are sometimes "unlabeled"

---

### Added
- Added the idp name to Sentry issues as a tag. This is searchable, sortable, and more visible than a context entry.

### Fixed
- Log a useful message to Sentry in cases where no explicit "message" key is included
- Fixed one specific error instance that had neither "message" nor "error"

### Changed
- Don't report user-generated error messages (i.e. http status codes 400, 401, 404, and 409)

---

### PR Checklist
- [x] Run `make composershow`
- [x] Run `make psr2`

### DIFF
--- a/application/dependencies.json
+++ b/application/dependencies.json
@@ -13,7 +13,7 @@
   },
   {
     "name": "clue/stream-filter",
-    "version": "v1.7.0"
+    "version": "1.x-dev b27b3b4"
   },
   {
     "name": "codemix/yii2-streamlog",
@@ -81,7 +81,7 @@
   },
   {
     "name": "jean85/pretty-package-versions",
-    "version": "2.0.6"
+    "version": "2.x-dev 3c4e5f6"
   },
   {
     "name": "mlocati/ip-lib",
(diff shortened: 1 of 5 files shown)


application/dependencies.json
```diff
@@ -13,7 +13,7 @@
   },
   {
     "name": "clue/stream-filter",
-    "version": "v1.7.0"
+    "version": "1.x-dev b27b3b4"
```
hobbitronics, line 16: I'm not familiar with this file or where these are coming from.

hobbitronics (approved):
Looks good, but I don't understand the dependencies.json file.

application/dependencies.json
briskt: Strange. I have no idea, but I'll do some digging.

briskt: I think they were changed to `dev` when I used `"minimum-stability": "dev"` temporarily. `composer install` didn't fix it so I did `composer remove silinternational/yii2-sentry` and `composer require silinternational/yii2-sentry` and that seemed to work.

forevermatt: @hobbitronics Just to pass along tribal knowledge, this `dependencies.json` file is created/updated by our `make composershow` command, so that diffs have a more human-friendly list of what dependencies and versions were updated by composer updates.

PR #393, opened 2025-01-14
Release 7.1.0 - Upgrade php version and dependencies

jason-jackson:
[IDP-941](https://itse.youtrack.cloud/issue/IDP-941) - Update to PHP 8.2 or later

---

### Changed
- Upgraded php to 8.3
- Updated dependencies

---

### PR Checklist
- [ ] Put version number in PR title (e.g. `Release x.y.z - Summary of changes`)
- [ ] Run `make composershow`
- [ ] Run `make psr2`


application/features/bootstrap/EmailContext.php
```diff
@@ -1168,7 +1168,7 @@ public function thatUserHasAPasswordThatExpiresInDays($n)
 
         $currentPassword = $this->tempUser->currentPassword;
         $currentPassword->password = base64_encode(random_bytes(33)); // Needed to pass validation
-        $currentPassword->expires_on = MySqlDateTime::relative('+' . $n . ' days');
+        $currentPassword->expires_on = MySqlDateTime::relative($n . ' days');
```
jason-jackson, line 1171: This fixed an email test where it was being passed `-1` and strtotime started to be unable to handle `+-1days` in our php8.3, but it worked in our php8.1.

application/common/helpers/MySqlDateTime.php
```diff
@@ -35,6 +35,7 @@ public static function formatDateTime(int $timestamp): string
      */
     public static function relative(string $difference = '+30 days'): string
     {
+        $difference = str_replace('--', '+', $difference);
```
jason-jackson, line 38: This fixed an email test where it was being passed `-2 days` and strtotime started to be unable to handle `--2 days` in our php8.3, but it worked in our php8.1.

application/features/bootstrap/EmailContext.php
briskt: tricky...good fix!

briskt (approved):
I'd like to know if there's a better fix for the `--` bug, but not if it's excessively difficult.

application/common/helpers/MySqlDateTime.php
jason-jackson: I've tried adjusting it in the [test](https://github.com/silinternational/idp-id-broker/blob/main/application/features/bootstrap/UnitTestsContext.php#L468), but nothing worked. I thought about adjusting it [here](https://github.com/silinternational/idp-id-broker/blob/main/application/common/models/User.php#L1398), but what I did now seemed a better place to fix it.

briskt: OK, I guess it can stay this way. Thanks.

jason-jackson: Though I wonder if it would be better to do a check to make sure it begins with the `--` instead of just a complete str_replace.

```
if (substr($difference, 0, 2) == '--') {
    $difference = substr($difference, 2);
} 
```

briskt: Or use trim instead of replace

jason-jackson: `trim` I think only is characters, not strings.

mtompset: It all starts with the app parameter being the negative of what it should be, because there's a weird logical relationship between the name of the parameter and how it is used. Perhaps an sprintf of the * -1 would be better?

briskt:
@jason-jackson is there a problem with this that it can't be merged?

PR #394, opened 2025-02-10
Release 7.0.2 -- reduce memory consumption of password expiry processing

briskt:
[IDP-1414](https://itse.youtrack.cloud/issue/IDP-1414) IdP Export Sheet

---

### Fixed
- Optimized the database query that retrieves users for sending password expiry messages. This task runs before the Google Sheets export (and the external groups sync). The logs show that the password expiry task is sometimes the last to run and it does not finish.

### Changed
- Changed the order of scheduled task execution. Moved password expiry processing to last in case the fix doesn't work.
- 
---

### PR Checklist
- [ ] Run `make composershow`
- [x] Run `make psr2`

### DIFF
--- a/application/common/components/Emailer.php
+++ b/application/common/components/Emailer.php
@@ -776,35 +776,17 @@ class Emailer extends Component
             'status' => 'starting',
         ];
 
-        $users = \Yii::$app->getDb()->createCommand("SELECT u.*
-            FROM `user` u
-                JOIN `password` p ON u.id = p.user_id
-                LEFT JOIN `email_log` e ON u.id = e.user_id
-                    AND e.message_type = 'password-expiring'
-                    AND e.sent_utc >= CURRENT_DATE() - INTERVAL ? DAY
-            WHERE u.active = 'yes'
-                AND u.locked = 'no'
-                AND p.expires_on < CURRENT_DATE() + INTERVAL 15 DAY
-                AND p.expires_on >= CURRENT_DATE() # send a different message if expired
-                AND e.id IS NULL
-            GROUP BY u.id
-            HAVING COUNT(*) = 1;")
-            ->bindValue(1, $this->emailRepeatDelayDays)
-            ->queryAll();
+        $users = User::getUsersForEmail('password-expiring', $this->emailRepeatDelayDays);
 
         $this->logger->info(array_merge($logData, [
             'users' => count($users)
         ]));
 
         $numEmailsSent = 0;
-        foreach ($users as $u) {
-            $user = new User();
-            User::populateRecord($user, $u);
-
+        foreach ($users as $user) {
             /** @var Password $userPassword */
             $userPassword = $user->currentPassword;
             if ($userPassword) {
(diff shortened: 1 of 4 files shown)


application/common/components/Emailer.php
```diff
@@ -776,21 +776,38 @@ public function sendPasswordExpiringEmails()
...
+                    AND e.message_type = 'password-expiring'
+                    AND e.sent_utc >= CURRENT_DATE() - INTERVAL ? DAY
+            WHERE u.active = 'yes'
+                AND u.locked = 'no'
+                AND p.expires_on < CURRENT_DATE() + INTERVAL 15 DAY
+                AND p.expires_on >= CURRENT_DATE() # send a different message if expired
+                AND e.id IS NULL
+            GROUP BY u.id
+            HAVING COUNT(*) = 1;")
+            ->bindValue(1, $this->emailRepeatDelayDays)
```
mtompset, line 792: Is there an 'emailRepeatDelayDays' field in the user table? Does this create a calculated field?

mtompset: Oh wait. That "INTERVAL ? DAY" is what is being bound?

application/common/components/Emailer.php
```diff
@@ -776,21 +776,38 @@ public function sendPasswordExpiringEmails()
             'status' => 'starting',
         ];
 
-        $users = User::getActiveUnlockedUsers();
+        $users = \Yii::$app->getDb()->createCommand("SELECT u.*
```
mtompset, line 779: I'm pretty sure you could use a non-active query to do the same thing.
$query = new Query(); // Not ActiveQuery, because that is slower.
$users = $query
-> select("u.*")
->from("user u")
->join('password p'.... Sorry, not quite up to thinking it all out...
->one();

Also, this was a pretty function. Nice one liner. Now it is a grotesque mess. Extract this perhaps?

forevermatt, line 779: Another option for reducing memory usage would be to use Yii2's built in batching capability. Docs: https://www.yiiframework.com/doc/guide/2.0/en/db-active-record#data-in-batches

forevermatt: That could also help us keep the query logic in the User class, rather than putting it here in the Emailer component.

briskt: To each his own. I don't think SQL is "grotesque". But you're right that the function grew in complexity.

application/common/components/Emailer.php
briskt: Yep

application/common/components/Emailer.php
briskt: I considered moving it to the User class. That class is already very bloated, so I resisted the urge.

briskt: > Another option for reducing memory usage would be to use Yii2's built in batching capability. Docs: https://www.yiiframework.com/doc/guide/2.0/en/db-active-record#data-in-batches

Thanks for this tip!

PR #396, opened 2025-02-19
Release 7.4.0 -- integrate email queue from email-service

briskt:
### Added
- Integrate email queue from email-service ([IDP-1444](https://itse.youtrack.cloud/issue/IDP-1444))

---

### PR Checklist
- [x] Run `make composershow`
- [x] Run `make psr2`


application/common/models/Email.php
```diff
@@ -0,0 +1,284 @@
...
+     * @return int number of sent messages
+     */
+    public function send(): int
+    {
+        /*
+         * Try to send email or throw exception
+         */
+        $message = $this->getMessage();
+        if (!$message->send()) {
+            throw new Exception('Unable to send email', 1461011826);
```
forevermatt, line 78: I don't know it's worth making this change, but it might be beneficial to put a new timestamp in here, so that as we search our codebase for a given error code, only one line of code in a single repo will match. That could avoid confusion regarding which repo (email-service or idp-id-broker) is the source of the problem. There should be other ways to tell (such as which application's logs we see that error code in), but I just thought I'd mention it in case you also think it's worth updating.

forevermatt: Same with the other error codes copied from email-service.

application/common/models/Email.php
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    /**
+     * Attempt to send messages from queue
+     * @throws Exception
+     */
+    public static function sendQueuedEmail()
+    {
+        $log = [
+            'action' => 'email/sendQueuedEmail',
```
forevermatt, line 191: It might also be worth updating this, since it will presumably not be [the `actionSendQueuedEmail()` Controller Action](https://github.com/silinternational/email-service/blob/main/application/console/controllers/SendController.php#L13) calling this code now.

Then again, this function has that name, too, so the only part that doesn't quite fit any more is the `action` key in the log data. Feel free to proceed as you see fit.

application/common/models/Email.php
```diff
@@ -0,0 +1,284 @@
...
+            'subject',
+            'text_body',
+            'html_body',
+            'attempts_count',
+            'updated_at',
+            'created_at',
+            'error',
+            'send_after',
+        ];
+    }
```
forevermatt, line 274: This might not be needed any more. I _think_ this `fields()` function is for controlling what is included when an Active Record model is returned by an API controller action (e.g. as JSON in an HTTP response). I'm not sure of the implications of removing it, though, for what that's worth.

application/common/models/EmailBase.php
forevermatt, line 1: Did this regenerate identically here to what it was in email-service? I'm mostly just curious, though it's also a smoke-test to help ensure we got the table defined correctly in ID Broker's database.

forevermatt (approved):

application/common/config/main.php
```diff
@@ -32,17 +34,30 @@
...
+        'encryption' => 'ssl',
+    ];
+} else {
+    $mailerConfig['class'] = SesMailer::class;
+    $mailerConfig['awsRegion'] = Env::get('AWS_REGION', 'us-east-1');
+}
 
-// Re-retrieve the validIpRanges as an array.
-$emailServiceConfig['validIpRanges'] = Env::getArray('EMAIL_SERVICE_validIpRanges');
+$fromEmail         = Env::get('FROM_EMAIL', 'no_reply@example.com');
```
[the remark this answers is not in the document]
briskt: You make a good point. I was going for simplicity, but adding a test value wouldn't be bad.

application/common/models/Email.php
briskt: I may implement the API endpoints so idp-pw-api and/or idp-id-sync can use id-broker as an email-service. In that case, the `fields` method would be used again.

Thoughts?

application/common/models/EmailBase.php
briskt: Yes, except that I omitted the `error` column. It was only used in tests. Thanks for the suggestion.

application/common/models/Email.php
forevermatt: Sounds good. Seems worth keeping it then. 👍

application/common/mail/layouts/html.php
```diff
@@ -1,34 +1,33 @@
...
 
+$brandColor = Env::get('EMAIL_BRAND_COLOR', '');
+$logo = Env::get('EMAIL_BRAND_LOGO', '');
+$maxWidth = Env::get('EMAIL_MAX_WIDTH', '600px');
+/* @var $content string email contents */
+
 ?>
 <!DOCTYPE html>
-<html>
+<html lang="eng">
```
forevermatt, line 12: I _think_ this should be `en` or `en-US`, rather than `eng`

briskt: Good catch, thank you.

forevermatt:
@briskt Is this still waiting on anything, or can it be merged/released? It looks like another PR (#403) is potentially conflicting with this PR's version number.

PR #403, opened 2025-03-27
Release 7.3.0 - Added option for user to request code from recovery contact

ethancanne:
### [IDP-1482](https://itse.youtrack.cloud/issue/IDP-1482)

This PR introduces a new MFA recovery method using email, allowing users to designate a recovery contact email. This feature is designed to provide an alternative way for users to regain access to their accounts in case they lose access to their primary MFA methods.

**Key Changes:**

* **New MFA Type (`recovery`):**
    * Added a new MFA type, `recovery`, to represent the email-based recovery method.
* ~~**Recovery Email Field:**~~
    * ~~Introduced a `recovery_email` field in the `mfa` table to store the recovery contact's email address.~~
* **Email Sending Functionality:**
    * Implemented functionality to send recovery codes to the designated recovery email address.
* **Verification Process:**
    * Added a verification process to validate the recovery code provided by the user.
* **Email Templates:**
    * ~~Created new~~ _Modified existing "manager"_ email templates for sending recovery codes and instructions to the recovery contact.
* **Configuration:**
    * ~~Added new configuration parameters for BCC email addresses for recovery-related emails.~~
    * Deprecated `SUBJECT_FOR_MFA_MANAGER` and `SUBJECT_FOR_MFA_MANAGER_HELP` fields, replacing them with `SUBJECT_FOR_MFA_RECOVERY` and `SUBJECT_FOR_MFA_RECOVERY_HELP`, respectively. The previous environment variable names will still be checked (in at least this version) to allow for a graceful transition to the new environment variable names.
* **Tests:**
    * Added comprehensive tests/scenarios to ensure the functionality works as expected.
* **Code Cleanup:**
    * Performed code cleanup and refactoring for better maintainability.
    * A few files were changed to clean up code, as shown in the diff.
* ~~**Migrations:**~~
    * ~~Added database migrations to support the new MFA type and recovery email field.~~
* **Updated Emailer component:**
    * Added constants and subject properties for recovery emails.
* **Updated Mfa Model:**
    * Added functionality to create and verify recovery MFA types.
* **Updated User model:**
    * Added functionality to remove recovery MFA types.
* **Updated Cron controller:**
    * Added functionality to remove old recovery MFA records.

---

### Testing
- Comprehensive tests have been added to cover the new functionality, including:
  - Creating recovery MFA records.
  - Verifying recovery codes.
  - Sending recovery emails.
  - Deleting recovery MFA options.

---

### PR Checklist
- [x] Run `make composershow`
- [x] Run `make psr2`


application/common/mail/mfa-recovery-help.html.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+use yii\helpers\Html as yHtml;
```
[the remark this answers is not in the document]
jason-jackson: It was copy/pasted from `manager-recovery-help.html.php`, so you'd have to take it up with @briskt. (I assume he probably copied it from another mail template).

application/common/config/main.php
```diff
@@ -141,6 +142,9 @@
             'subjectForMfaDisabled' => Env::get('SUBJECT_FOR_MFA_DISABLED'),
             'subjectForMfaManager' => Env::get('SUBJECT_FOR_MFA_MANAGER'),
             'subjectForMfaManagerHelp' => Env::get('SUBJECT_FOR_MFA_MANAGER_HELP'),
+            'subjectForMfaRecovery' => Env::get('SUBJECT_FOR_MFA_RECOVERY'),
+            'subjectForMfaRecoveryHelp' => Env::get('SUBJECT_FOR_MFA_RECOVERY_HELP'),
```
briskt, line 146: Even though we've used environment variables for subject customization in the past, I don't think we should this time. For one, only one IdP instance is going to use this option. And we almost have never actually used custom subject text.

application/common/config/main.php
```diff
@@ -290,6 +295,8 @@
         'mfaLifetime'                   => Env::get('MFA_LIFETIME', '+2 hours'),
         'mfaManagerBcc'                 => Env::get('MFA_MANAGER_BCC', ''),
         'mfaManagerHelpBcc'             => Env::get('MFA_MANAGER_HELP_BCC', ''),
+        'mfaRecoveryBcc'                => Env::get('MFA_RECOVERY_BCC', ''),
+        'mfaRecoveryHelpBcc'            => Env::get('MFA_RECOVERY_HELP_BCC', ''),
```
briskt, line 299: Likewise, I think these aren't needed. From what I can recall, we never used them for the "Manager" option. My guess is we won't use it with the "Recovery" option either.

application/common/mail/mfa-recovery-help.html.php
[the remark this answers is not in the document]
briskt: @mtompset why, what's wrong with yHtml?

application/common/mail/mfa-recovery-help.html.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
briskt, line 1: Since this template is identical to `mfa-manager-help.html.php` except for the name of the `recoveryEmail` variable, could we just use one template for both and change the variable name to something that covers both scenarios?

application/common/mail/mfa-recovery.html.php
```diff
@@ -0,0 +1,46 @@
+<?php
```
briskt, line 1: Since this template is identical to `mfa-manager.html.php` is it necessary? Could we just use one template for both?

application/common/models/Mfa.php
```diff
@@ -334,10 +341,18 @@ public static function create(int $userId, string $type, string $label = null, s
...
+        }
+
         $existing = self::findOne(['user_id' => $userId, 'type' => $type, 'verified' => 1]);
 
         if ($existing instanceof Mfa) {
-            if ($type == self::TYPE_BACKUPCODE || $type == self::TYPE_MANAGER || $type == self::TYPE_WEBAUTHN) {
+            if ($type == self::TYPE_BACKUPCODE || $type == self::TYPE_MANAGER || $type == self::TYPE_RECOVERY || $type == self::TYPE_WEBAUTHN) {
                 $mfa = $existing;
             } else {
                 throw new ConflictHttpException('An MFA of type ' . $type . ' already exists.', 1551190694);
```
briskt, line 358: It's probably time to reverse this logic
```suggestion
            if ($type == self::TYPE_TOTP) {
                throw new ConflictHttpException('An MFA of type ' . self::TYPE_TOTP . ' already exists.', 1551190694);
            } else {
                $mfa = $existing;
```

briskt (changes requested):
Good work getting this all in place. As with the related changes in ssp-base, I don't like that the manager and recovery are nearly identical but treated as separate types. But I think we're past that decision point. Nevertheless, there are ways that some duplication can be avoided. I've made some suggestions in the code.

application/common/mail/mfa-recovery-help.html.php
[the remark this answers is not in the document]
mtompset: It's nothing worthy holding things up for.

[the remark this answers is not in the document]
briskt: (From DM conversation:) I see now that the alias is unnecessary. It's a very historic thing in many files in this repo. Better to leave it as it is in this PR, change all occurrences at once in a separate PR, or just let it go.

application/common/config/main.php
forevermatt: I agree 👍

application/features/mfa.feature
```diff
@@ -62,6 +62,43 @@ Feature: MFA
...
+    When I request "/mfa" be created
+    Then the response status code should be 400
+
+  Scenario: Create new MFA record of type recovery with an invalid recovery email
+    Given I provide the following valid data:
+      | property       | value        |
+      | employee_id    | 123          |
+      | type           | recovery     |
+      | label          | A Label      |
+      | recovery_email | invalidEmail |
```
forevermatt, line 98: The wording of this test seems contradictory: "the following valid data... invalidEmail".

We might have a wording "bug" in our behat tests for ID Broker. However, it looks like other tests already have this problem, so this is likely something to fix in a later PR. If you would, please make a note (possibly in this same YouTrack issue) to come back and look at this later.

application/common/models/Mfa.php
```diff
@@ -346,6 +361,7 @@ public static function create(int $userId, string $type, string $label = null, s
             $mfa = new Mfa();
             $mfa->user_id = $userId;
             $mfa->type = $type;
+            $mfa->recovery_email = $recoveryEmail;
```
forevermatt, line 364: I don't fully understand: why do we need to save the recovery email address to the database? It seems like we could simply pass it to regInit (below, on new line 384), like we do with `rpOrigin`.

Feel free to fill me in on your thinking here, as I expect you have some meaningful reason for having done it this way. (Thanks)

application/common/models/Mfa.php
forevermatt: You could either do that or extract a method named for what question this is trying to ask. For example, `isTypeThatUsesChildRecords($type)` or something like that. I'm not sure if that's the exact question this if-condition is trying to ask, but I think it's something close to that. Feel free to adjust wording as you see fit if you choose to go this route.

I slightly prefer a whitelisted approach (= it's one of these) rather than blacklisted (it's not TOTP), but I also respect @briskt's input on this. It has certainly gotten too long for a single if-condition at this point.

application/common/models/Mfa.php
```diff
@@ -493,6 +509,7 @@ public static function getTypes()
             self::TYPE_MANAGER => 'Manager Backup Code',
             self::TYPE_TOTP => 'Authenticator App',
             self::TYPE_WEBAUTHN => 'Security Key',
+            self::TYPE_RECOVERY => 'Recovery Backup Code',
```
forevermatt, line 512: It might be better to call this a "Recovery Contact Code", since it's for use by a recovery contact to help the person recover from having lost their MFA methods:
```suggestion
            self::TYPE_RECOVERY => 'Recovery Contact Code',
```

I think we used "Backup" in the manager option to communicate that concept of helping someone get back into their account, but "recovery backup" doesn't quite communicate our intended purpose, I think.

Feel free to speak up, if you see it differently.

application/console/controllers/CronController.php
```diff
@@ -21,6 +21,9 @@ public function actionRemoveOldUnverifiedRecords()
         Mfa::removeOldUnverifiedRecords();
 
         Mfa::removeOldManagerMfaRecords();
+
+        Mfa::removeOldRecoveryMfaRecords();
+
```
forevermatt, line 26: Minor cleanup (to avoid blank line as the last line of a function):

```suggestion
```

forevermatt (commented):
Partial progress reviewing, not done yet.

application/features/bootstrap/EmailContext.php
```diff
@@ -1059,6 +1061,14 @@ public function aManagerRescueEmailIsSentToTheManager()
         $this->assertEmailSent(EmailLog::MESSAGE_TYPE_MFA_MANAGER, self::MANAGER_EMAIL);
     }
 
+    /**
+     * @Then a Recovery Rescue email is sent to the recovery contact
+     */
+    public function aRecoveryRescueEmailIsSentToTheRecoveryContact()
+    {
+        $this->assertEmailSent(EmailLog::MESSAGE_TYPE_MFA_RECOVERY, $this->testMfaOption->recovery_email);
```
forevermatt, line 1069: It seems like you could simplify this to just use the constant (like was done with the manager rescue email):
```suggestion
        $this->assertEmailSent(EmailLog::MESSAGE_TYPE_MFA_RECOVERY, self::RECOVERY_EMAIL);
```
Was there a benefit to using the `testMfaOption` test property instead?

application/common/models/Mfa.php
jason-jackson: Good point, we made the create first, and for some reason I was thinking we'd need it for verification, but we don't. we can remove this.

application/common/models/Mfa.php
jason-jackson: True, we could easily use a switch here.

briskt: A switch isn't what either of us recommended. It would only make a very long function longer.

application/features/bootstrap/EmailContext.php
jason-jackson: testMfaOption was only there because we had it stored there, since we're removing that from being stored, it would be better to do what you're suggesting.

application/console/migrations/m250318_201324_add_mfa_type_recovery.php
```diff
@@ -0,0 +1,62 @@
...
+     * {@inheritdoc}
+     */
+    public function safeDown()
+    {
+        // remove recovery mfa type
+        $this->alterColumn(
+            '{{mfa}}',
+            'type',
+            "enum('totp','u2f','backupcode','manager','webauthn') not null"
+        );
```
jason-jackson, line 48: Since we're changing enum values, we'd need to do something about stored values *before* we alter the column. I think deleting them is fine.

application/console/migrations/m250318_201324_add_mfa_type_recovery.php
```diff
@@ -0,0 +1,62 @@
...
+
+        // Drop column that stores recovery email
+        $this->dropColumn(
+            '{{mfa}}',
+            'recovery_email'
+        );
+
+        $this->alterColumn(
```
jason-jackson, line 56: Since we're changing enum values, we'd need to do something about stored values before we alter the column.

application/common/mail/mfa-recovery-help.html.php
ethancanne: I agree, we can have a single mfa-recovery email template that works for sending codes to both the manager and recovery contact.

forevermatt:
I'm ready to test this out on our staging IDP. Are there more changes you are planning to make to this PR? If there is something you  are waiting on any of us for, feel free to say so. Otherwise I'll just wait to hear from you about when this is ready/released. (Thanks)

ethancanne:
@forevermatt, I just pushed several changes that address the suggestions you and @briskt made.

application/common/components/MfaBackendInterface.php
```diff
@@ -9,9 +9,10 @@ interface MfaBackendInterface
      * @param int $userId
      * @param string $mfaExternalUuid (only used for WebAuthn)
      * @param string $rpOrigin Relying Party Origin (only used for WebAuthn)
+     * @param string $recoveryEmail for mfa account recovery (only used for WebAuthn)
      * @return array
      */
-    public function regInit(int $userId, string $mfaExternalUuid, string $rpOrigin): array;
+    public function regInit(int $userId, string $mfaExternalUuid, string $rpOrigin, ?string $recoveryEmail = null): array;
```
forevermatt, line 15: This is possibly too nit-picky, but it may be better to follow the example of the `$rpOrigin` parameter, and default to an empty string instead of making `$recoveryEmail` a nullable string:
```suggestion
    public function regInit(int $userId, string $mfaExternalUuid, string $rpOrigin, string $recoveryEmail = ''): array;
```

application/common/config/main.php
forevermatt: I don't know if it's worth it, but if we want to maintain backwards compatibility for this, I think we would need to fallback to the `SUBJECT_FOR_MFA_MANAGER` / `SUBJECT_FOR_MFA_MANAGER_HELP` env. vars.
```suggestion
            'subjectForMfaRecovery' => Env::get(
                'SUBJECT_FOR_MFA_RECOVERY',
                Env::get('SUBJECT_FOR_MFA_MANAGER')
            ),
            'subjectForMfaRecoveryHelp' => Env::get(
                'SUBJECT_FOR_MFA_RECOVERY_HELP',
                Env::get('SUBJECT_FOR_MFA_MANAGER_HELP')
            ),
```
Again, I don't know if it's worth this, especially given that we are moving away from these customizable subject lines (due to complexity and lack of use).

application/common/models/Mfa.php
```diff
@@ -308,12 +312,14 @@ public function verify($value, string $rpOrigin = '', string $verifyType = '', s
...
      * @param string|null $label
+     * @param string $rpOrigin
+     * @param string|null $recoveryEmail
      * @return array
      * @throws BadRequestHttpException
      * @throws ServerErrorHttpException
      * @throws ConflictHttpException
      */
-    public static function create(int $userId, string $type, string $label = null, string $rpOrigin = ''): array
+    public static function create(int $userId, string $type, ?string $label = null, string $rpOrigin = '', ?string $recoveryEmail = ''): array
```
forevermatt, line 322: Since this is defaulting to an empty string, I suspect it doesn't need to be nullable
```suggestion
    public static function create(int $userId, string $type, ?string $label = null, string $rpOrigin = '', string $recoveryEmail = ''): array
```

application/console/migrations/m250318_201324_add_mfa_type_recovery.php
```diff
@@ -0,0 +1,71 @@
...
+            '{{email_log}}',
+            'message_type',
+            "enum('invite','welcome','mfa-rate-limit','password-changed','get-backup-codes','refresh-backup-codes','lost-security-key','mfa-option-added','mfa-option-removed','mfa-enabled','mfa-disabled','method-verify','mfa-manager','mfa-manager-help','method-reminder','method-purged','password-expiring','password-expired','password-pwned','ext-group-sync-errors','abandoned-users', 'mfa-recovery', 'mfa-recovery-help') null"
+        );
+
+        //Updating all manager email records to recovery emails
+        $this->update("{{email_log}}", ["message_type" => "mfa-recovery"], ["message_type" => "mfa-manager"]);
+
+        $this->update("{{email_log}}", ["message_type" => "mfa-recovery-help"], ["message_type" => "mfa-manager-help"]);
```
forevermatt, line 34: I hesitate to modify history records, though I think what you did here will work. I just wonder if it's better to leave those records untouched.

Actually, perhaps your approach is better, since going forward we will refer to these as mfa-recovery(-help) emails, and the EmailLog is for tracking when/whether we have already sent such an email. Leaving the old records as "manager" could result in another email being sent (since the EmailLog won't have a record that a "recovery" email was sent).

Okay, my rambling here is done. No need to change your approach here after all.

application/features/bootstrap/EmailContext.php
```diff
@@ -1193,35 +1211,19 @@ public function iSendPasswordExpiredEmails()
     }
 
     /**
-     * @Given a mfaManagerBcc email address is configured
-     */
-    public function aMfamanagerbccEmailAddressIsConfigured()
-    {
-        \Yii::$app->params['mfaManagerBcc'] = 'email@example.com';
```
forevermatt, line 1200: One of our IDPs _might_ be using this "bcc" feature. I'm checking on that...

application/features/bootstrap/EmailContext.php
```diff
@@ -1193,35 +1211,19 @@ public function iSendPasswordExpiredEmails()
...
-    }
-
-    /**
-     * @Then the mfaManagerBcc email address is on the bcc line
-     */
-    public function theMfamanagerbccEmailAddressIsOnTheBccLine()
```
forevermatt, line 1206: This test step is still in use on line 328 of `application/features/email.feature`, so we probably shouldn't delete it here (unless we also delete it there... but I think that decision might still be pending).

forevermatt:
> @forevermatt, I just pushed several changes that address the suggestions you and @briskt made.

Thanks, @ethancanne (and @jason-jackson)! I think this is getting very close to done... just a few more things to consider. Sorry this has had so much back-and-forth, but good work.

briskt (approved):

application/common/components/MfaBackendInterface.php
jason-jackson: I don't have a strong preference either way. We did it this way because the getParam can return null, but that is easily fixable.

application/common/models/Mfa.php
jason-jackson: Unless we change the controller up, it should be nullable. `$recoveryEmail = $req->getBodyParam('recovery_email');` is set to null if it isn't provided.

application/features/bootstrap/EmailContext.php
jason-jackson: Looks like it is.

jason-jackson: This is fine though as it is replaced by `aParamEmailAddressIsConfigured`, we could roll it back as it only replaces two functions instead of four.

application/features/email.feature
```diff
@@ -326,15 +326,27 @@ Feature: Email
     When I request a new manager mfa
     Then a Manager Rescue email is sent to the manager
       And the mfaManagerBcc email address is on the bcc line
-
+
```
jason-jackson, line 329:
```suggestion

```

local.env.dist
```diff
@@ -215,6 +215,14 @@ HR_NOTIFICATIONS_EMAIL=test@test.com
...
 #MFA_MANAGER_HELP_BCC=
 
+# Email address to include in bcc for recovery mfa emails
+# By default does not include a bcc address.
+#MFA_RECOVERY_BCC=
+
+# Email address to include in bcc for recovery mfa help emails
+# By default does not include a bcc address.
+#MFA_RECOVERY_HELP_BCC=
+
```
jason-jackson, line 225:
```suggestion
```

forevermatt: It appears that one of our IDPs _is_ using this feature after all, so we'll need to keep some way for recovery codes (whether sent to the manager or an alternate recovery contact) to also be bcc'd to another email address.

application/common/models/Mfa.php
forevermatt: We can easily change that to be an empty string instead of null. I think avoiding null (except when we _really_ need it) will benefit us in the long run.

But, this is almost splitting hairs. Feel free to proceed as seems best to you two on this issue.

local.env.dist
ethancanne: Since one of the IDPs is still using this feature, should these values be removed from the `local.env.dist`?

briskt: The only IdP that uses the manager code bcc feature is the usa IdP, which isn't going to need the new recovery code feature set. So I think `MFA_RECOVERY_BCC` and `MFA_RECOVERY_HELP_BCC` are not needed.

forevermatt (approved):
Good to go. I updated the description slightly, to reflect some of the changes made during the PR review.

application/frontend/controllers/MfaController.php
```diff
@@ -45,13 +45,15 @@ public function actionCreate()
 
         $label = $req->getBodyParam('label');
 
+        $recoveryEmail = $req->getBodyParam('recovery_email') ?? '';
```
jason-jackson, line 48:
```suggestion
        $recoveryEmail = $req->getBodyParam('recovery_email', '');
```

application/common/components/MfaBackendTotp.php
```diff
@@ -43,7 +43,7 @@ public function init()
         parent::init();
     }
 
-    public function regInit(int $userId, string $mfaExternalUuid = null, string $rpOrigin = ''): array
+    public function regInit(int $userId, string $mfaExternalUuid = null, string $rpOrigin = '', ?string $recoveryEmail = null): array
```
jason-jackson, line 46:
```suggestion
    public function regInit(int $userId, string $mfaExternalUuid = null, string $rpOrigin = '', string $recoveryEmail = ''): array
```

application/common/components/MfaBackendInterface.php
```diff
@@ -9,9 +9,10 @@ interface MfaBackendInterface
      * @param int $userId
      * @param string $mfaExternalUuid (only used for WebAuthn)
      * @param string $rpOrigin Relying Party Origin (only used for WebAuthn)
+     * @param string $recoveryEmail for mfa account recovery (only used for WebAuthn)
      * @return array
      */
-    public function regInit(int $userId, string $mfaExternalUuid, string $rpOrigin): array;
+    public function regInit(int $userId, string $mfaExternalUuid, string $rpOrigin, string $recoveryEmail = ""): array;
```
jason-jackson, line 15:
```suggestion
    public function regInit(int $userId, string $mfaExternalUuid, string $rpOrigin, string $recoveryEmail): array;
```

PR #409, opened 2025-05-12
Release 8.0.4 -- (3rd try) fix the email header style

briskt:
### Fixed
- move text-align to the th element


hobbitronics (approved):
Third tries a charm. Are you not able to inspect the css in the email to see what works?

briskt:
> Third tries a charm. Are you not able to inspect the css in the email to see what works?

I haven't tried it locally. Each time, I thought it was a simple change and didn't need to be checked in a local instance of idp-id-broker. I did manually change the CSS in the browser, but I must have not done it the same way in the code.

PR #412, opened 2025-07-16
fix password_verify deprecation by checking for null first

briskt:
### Fixed
- Fix deprecation in `password_verify` by checking for a null first.

### DIFF
--- a/application/common/models/User.php
+++ b/application/common/models/User.php
@@ -292,7 +292,8 @@ class User extends UserBase
     {
         return function ($attributeName) {
             $currentPassword = $this->currentPassword ?? new Password();
-            if ($currentPassword->hash === null || !password_verify($this->password, $currentPassword->hash)) {
+            $hash = $currentPassword->hash ?? null;
+            if ($hash === null || !password_verify($this->password, $currentPassword->hash)) {
                 $this->addError($attributeName, 'Incorrect password.');
             } else {
                 // check the current hash cost and rehash if necessary


application/common/models/User.php
```diff
@@ -292,7 +292,7 @@ private function validatePassword(): Closure
     {
         return function ($attributeName) {
             $currentPassword = $this->currentPassword ?? new Password();
```
mtompset, line 294: Where is $this->currentPassword defined? PHP 8.x doesn't allow automatic creation of properties.

mtompset: Yii 2.x relationship defined in UserBase.

application/common/models/User.php
```diff
@@ -292,7 +292,7 @@ private function validatePassword(): Closure
...
-            if (!password_verify($this->password, $currentPassword->hash)) {
+            if ($currentPassword->hash === null || !password_verify($this->password, $currentPassword->hash)) {
```
mtompset, line 295: Are you sure you can access hash on new Password()? I would have ?? null, and then ?->hash.

briskt: No, you'd know better than I. However, this has been working for years until now that PHP is complaining about the second parameter of `password_verify` not accepting null.

mtompset: I'm just worried new Password() may not initialize hash and then your check might fail. This is why I suggested the change to 294 (?? null) and then the new check (?->hash).

mtompset: If you can tell me the line of code where a `new Password()` initializes the `hash` property to `null`, then as is is okay. Otherwise, I'd recommend the coding change I suggested. PHP 8.x is finicky about accessing variables before they are initialized.

briskt: Is this adequate? 
```php
    public function rules(): array
    {
        return ArrayHelper::merge([
// ...
            [
                'hash', 'default', 'value' => function () {
                    return self::hashPassword($this->password);
                },
            ],
// ...

        ], parent::rules());
    }
```

https://github.com/silinternational/idp-id-broker/blob/3ac20ea34735c529e5bed123769467e6ac875d10/application/common/models/Password.php#L56

mtompset: The problem is rules are generally only triggered on validation. There's no validation between new Password() and the logic below it. Models have their init() called on construction.

briskt: OK, thanks. I tried it in PHP interactive shell, and the `?->` didn't seem to help. So I just added `?? null`.

PR #413, opened 2025-07-21
Release 8.2.0 -- database SSL connection

briskt:
[IDP-1598](https://support.gtis.sil.org/issue/IDP-1598) Update idp-id-broker to talk with RDS over SSL

---

### Added
- Load the AWS RDS certificate bundle from an environment variable
- Added test key/cert to the database to verify database SSL transport is functional

---

### PR Checklist
- [x] Run `make composershow`
- [x] Run `make psr2`


docker-compose.yml
```diff
@@ -18,6 +21,7 @@ services:
       - db
     env_file:
       - ./local.env
+      - ./db/certs/test.env
```
mtompset, line 24: why not the same path/required logic for local.env here?

application/run-cron.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+# exit if any line in the script fails
+set -e
```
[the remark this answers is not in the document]
briskt: Good suggestion.

docker-compose.yml
briskt: It's only necessary for files that are git-ignored

PR #414, opened 2025-07-21
remove actions-services.yml

briskt:
### Removed
- Removed the actions-services.yml file and use docker-services.yml for all testing


docker-compose.yml
```diff
@@ -7,7 +7,8 @@ services:
     ports:
       - "3306"
     env_file:
-      - ./local.env
+      - path: ./local.env
```
mtompset, line 10: with the addition of cert stuff, are the user name, password, etc. in local.env no longer needed? If so, shouldn't the local.env.dist be edited too?

mtompset (approved):

PR #451, opened 2026-04-09
Remove whenavail

briskt:
[ITSE-1081](https://support.gtis.sil.org/issue/ITSE-1081) where possible, replace whenavail with health checks

---

### Removed
- Removed all use of [whenavail](https://github.com/sil-org/whenavail-script) and use docker health checks instead.


hobbitronics (approved):
Looks good

PR #453, opened 2026-04-17
Fix the cc and bcc headers in emails sent. The SES driver was incomplete

briskt:
[IDP-1949](https://support.gtis.sil.org/issue/IDP-1949) Add cc option to invite email

---

### Fixed
- Fixed a bug that left the cc and bcc headers empty when sending through AWS SES.

---

### PR Checklist
- [ ] Run `make composershow`
- [x] Run `make psr2`

### DIFF
--- a/app/common/components/SesMailer.php
+++ b/app/common/components/SesMailer.php
@@ -50,12 +50,12 @@ class SesMailer extends BaseMailer
         ];
 
         $cc = $message->getCc();
-        if (!empty($cc)) {
+        if (!empty($cc) && !empty($cc[0])) {
             $destination['CcAddresses'] = $cc;
         }
 
         $bcc = $message->getBcc();
-        if (!empty($bcc)) {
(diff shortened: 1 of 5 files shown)


app/common/components/SesMailer.php
```diff
@@ -45,11 +45,23 @@ public function init()
      */
     protected function sendMessage($message)
     {
+        $destination = [
+            'ToAddresses' => $message->getTo(),
+        ];
+
+        $cc = $message->getCc();
```
jason-jackson, line 52: Should getCc and getBcc incorporate a self::NONE check too? Or does mailer know how to parse `(none)`?

jason-jackson (approved):

app/common/components/SesMailer.php
briskt: The self::NONE is only used for the log. It's not part of the email API.

PR #454, opened 2026-04-17
Update the SES mail driver to more correctly match the Yii2 interface.

briskt:
[IDP-1949](https://support.gtis.sil.org/issue/IDP-1949) Add cc option to invite email

---
### Added
- Added unit tests for the SesMessage and SesMailer classes. (Sorry for the huge commit. Behat is very wordy.)

### Fixed
- Update the SES mail driver to more correctly match the Yii2 interface.
- Don't use the From address as the Reply-To. Aside from being unnecessary, the Reply-To is a list and the From is a single address, which causes a PHP type error if not dealt with.

---

### PR Checklist
- [ ] Run `make composershow`
- [x] Run `make psr2`

### DIFF
--- a/app/common/components/SesMessage.php
+++ b/app/common/components/SesMessage.php
@@ -108,10 +108,7 @@ class SesMessage extends BaseMessage
      */
     public function getReplyTo(): string|array
     {
-        if ($this->replyTo) {
-            return $this->replyTo;
-        }
-        return $this->getFrom();
+        return $this->replyTo;
     }
 
     /**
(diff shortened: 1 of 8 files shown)


app/common/components/SesMessage.php
```diff
@@ -13,48 +14,61 @@
 class SesMessage extends BaseMessage
 {
     /** @var string */
-    private $charset;
+    private string $charset = '';
```
mtompset, line 17: Your preference for strong typing is showing, Go coder. ;)

mtompset (approved):

PR #457, opened 2026-04-24
update dependencies.json for dependabot patch

briskt:
### Fixed
- Updated dependencies.json for the dependabot patch in #452

### DIFF
--- a/app/dependencies.json
+++ b/app/dependencies.json
@@ -133,7 +133,7 @@
   },
   {
     "name": "phpseclib/phpseclib",
-    "version": "3.0.50"
+    "version": "3.0.51"
   },
   {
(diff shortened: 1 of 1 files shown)


mtompset (approved):

mtompset:
No other dependencies to update while doing this?

briskt:
> No other dependencies to update while doing this?

I think the update philosophy here is different than you have for Mail Admin. Here we update only as required.

briskt:
> > No other dependencies to update while doing this?
> 
> I think the update philosophy here is different than you have for Mail Admin. Here we update only as required.

And this isn't actually an update, but a catch-up on this generated file since dependabot doesn't know to do that.

PR #460, opened 2026-04-27
rebuild base models using the current version of Yii2

briskt:
### Fixed
- Rebuild base models using the rebuildbasemodels.sh script on the current version of Yii2.


briskt:
This breaks things. We should probably abandon this framework.

PR #462, opened 2026-05-07
deprecate the "hide" property

briskt:
[IDP-1996](https://support.gtis.sil.org/issue/IDP-1996) Remove "hide" property from idp-id-broker

---

### Deprecated
- Deprecate the "hide" property. It is intended to be always-on in the future.

---

### PR Checklist
- [ ] Run `make composershow`
- [ ] Run `make psr2`


briskt:
Converting the spec to OpenAPI before the deprecation.

PR #463, opened 2026-05-07
upgrade API doc to OpenAPI 3

briskt:
### Changed (non-breaking)
- Upgraded the API spec document from RAML to OpenAPI 3. Disclaimer: I used an automated conversion tool, so the result may not be 100% accurate. I did check for accuracy, though, and it seems to be correct. I didn't replace the raml2html utility because OpenAPI rendering is a common feature in IDEs and IDE plugins.

---

### PR Checklist
- [ ] Run `make composershow`
- [ ] Run `make psr2`


mtompset (approved):
Good thing I never really started to learn raml. ;)

briskt:
> Good thing I never really started to learn raml. ;)

It was simpler to learn, almost human-readable, compared to OpenAPI. But the IDE-rendered OpenAPI is much, much better.

PR #470, opened 2026-05-19
include an option to send a password reset to the user's manager

briskt:
[IDP-2033](https://support.gtis.sil.org/issue/IDP-2033) API endpoint to initiate password reset through user's supervisor

---

### Added
- Include an option to send a password reset to the user's manager.

---

### PR Checklist
- [x] Documentation (README, local.env.dist, openapi.json, etc.)
- [x] Run `make composershow`
- [x] Run `make psr2`


app/common/models/Reset.php
```diff
@@ -128,21 +128,18 @@ public function isExpired(): bool
...
-    protected function send(): void
+    protected function send(bool $includeManager): void
     {
-        Yii::info("sending reset to employee '{$this->user->employee_id}' primary email: " . $this->user->email);
         $this->sendPrimary();
 
         $methods = $this->user->getVerifiedMethodOptions();
-        if (empty($methods) && !empty($this->user->manager_email)) {
-            Yii::info("sending reset to employee '{$this->user->employee_id}' manager: {$this->user->manager_email}");
+        if (empty($methods) && !empty($this->user->manager_email) || $includeManager) {
```
mtompset, line 139: Is this condition is wrong?
(empty() || $include) && !empty()?

mtompset: Oh wait... the empty methods is attempt to optionally send it to manager, but $include is actually send it. Okay, the exception in the sendManager() for a $include with no manager email makes sense.

app/features/reset-unit-tests.feature
```diff
@@ -46,6 +49,15 @@ Feature: Password Reset
     And the reset record has an expiry in the future
     And a "reset-self" email should be sent to their primary email
 
+  Scenario: Create a password reset with the "include manager" option
+    Given there is a user in the database with a valid password reset
+    And that user has a "manager_email" property value of "manager@example.com"
```
mtompset, line 54: Should add a test for the no manager email case which triggers an exception.

briskt: There's a no-manager scenario on line 22: "Scenario: Create a password reset for a user with no recovery options and no manager". You're not talking about a PHP exception, right (i.e. `throw new Exception`)?

hobbitronics (approved):
Looks good

PR #477, opened 2026-06-01
IDP-1804 - CC mail admins on invite user emails

samuel-sowpati-sil:
[IDP-1804](https://support.gtis.sil.org/issue/IDP-1804) CC Mail Admin on IdP Onboarding

---

### Added
- `ApiClient.php`: Guzzle client for authenticated (Bearer) GET calls.
- `MailAdmin.php`: `getEmailsFor()` fetches an account's mail admins from the Mail Admin API.
- `User::getInviteCcAddress()`: CCs an account's mail admins (or fallback) on the new-user invite, gated by `ACCOUNT_MAIL_ADMINS_CC_ON_INVITE` (default off).
- Four `ACCOUNT_MAIL_ADMINS_*` config params in `main.php`.
- `mail-admin` (Mockoon) service + mock data, with a healthcheck.
- Behat scenarios: CC on/off, personal+admin CC, empty-list fallback, no-fallback; `Email` comma-separated CC split + validation.

### Changed (non-breaking)
- `Email::getMessage()`: splits a comma-separated `cc_address` into individual recipients.

### Fixed
- `Email` validation: `cc_address` now accepts a comma-separated list (was single-address only), so multi-recipient CC no longer fails to queue.

---


app/common/components/MailAdmin.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        $apiUrl = Yii::$app->params['userMailAdminsApiUrl'] ?? '';
+        $apiKey = Yii::$app->params['userMailAdminsApiKey'] ?? '';
+
+        if (empty($apiUrl) || empty($accountEmail)) {
+            return [];
+        }
+
+        try {
+            $mailAdmins = (new ApiClient($apiKey))->call($apiUrl, ['email' => $accountEmail]);
```
briskt, line 30: Can `ApiClient` be combined into `MailAdmin`? Both are pretty small and are single-purpose. I'm not sure it's helpful to have the separation in this particular case.

Alternatively, a follow-up PR could refactor other uses of `GuzzleHttp\Client` in this project to use this new `ApiClient` wrapper. But even that doesn't seem like a significant advantage.

app/common/components/SesMessage.php
```diff
@@ -268,7 +268,9 @@ public function toString(): string
      */
     private static function formatAddresses(string|array $email): array
     {
-        $email = is_array($email) ? $email : [$email];
+        if (is_string($email)) {
```
briskt: `SesMessage` inherits from the Yii2 `BaseMessage` class. In that class, email address properties should be either a string, for a single address, or an array for multiple. Instead of passing a comma-separated list, can you use an array of individual addresses?

app/features/mocks/account-mail-admins.json
```diff
@@ -0,0 +1,121 @@
+{
```
briskt, line 1: I would prefer to see this in the `features` directory (i.e. `features/mocks`) to keep test-related items grouped together.

app/common/config/main.php
```diff
@@ -334,5 +334,9 @@
         'passwordProfileUrl'            => $passwordProfileUrl . "/#",
         'passwordReuseLimit'            => Env::get('PASSWORD_REUSE_LIMIT', 10),
         'profileReviewInterval'         => Env::get('PROFILE_REVIEW_INTERVAL', '+6 months'),
+        'userMailAdminsApiKey'          => Env::get('USER_MAIL_ADMINS_API_KEY'),
```
briskt: Since the MailAdmin API could theoretically be used for other purposes, what would you think of making the variable name prefix a bit more inclusive? Something like `MAIL_ADMIN_API_KEY`, perhaps?

Note: I did look at the ssp-base implementation and found that the variable name prefix is very specific. This highlights the problem. It would be nice to piggyback on those same variables since they're defined in the same namespace as these will be. But to do so would mean that these variables would have a prefix of `MFA_RECOVERY_CONTACTS` which would be incorrect here.

app/common/components/MailAdmin.php
```diff
@@ -0,0 +1,46 @@
...
+     * Returns an empty array when the feature is not configured or when the call
+     * to Mail Admin fails, so that a Mail Admin outage never blocks the email that
+     * triggered the lookup.
+     *
+     * @param string $accountEmail The email address of the account whose mail admins to fetch.
+     * @return string[] The mail admins' email addresses.
+     */
+    public static function getEmailsFor(string $accountEmail): array
+    {
+        $apiUrl = Yii::$app->params['userMailAdminsApiUrl'] ?? '';
```
briskt: Another reason to drop the word "user" from the configuration: the class name `MailAdmin` doesn't have the word "user" in it.

app/common/models/User.php
```diff
@@ -804,6 +805,22 @@ protected function sendAppropriateMessages($isNewUser, $changedAttributes)
             if ($this->personal_email && $this->email) {
                 $data['ccAddress'] = $this->personal_email;
             }
+            if (\Yii::$app->params['userMailAdminsCcOnInvite']) {
+                $mailAdminEmails = MailAdmin::getEmailsFor($this->email);
+                if (empty($mailAdminEmails)) {
+                    $fallback = \Yii::$app->params['userMailAdminsCcFallback'] ?? '';
+                    if (!empty($fallback)) {
```
briskt: Four levels of conditional statements is too much. (Three is also arguably too much.) Please refactor this.

As you do, consider this: I nearly suggested that the `if (!empty($mailAdminEmails))` line could be replaced with `else` until I realized that `$mailAdminEmails` is modified in the first statement. Maybe there's a way to make it easier to see that detail.

app/features/bootstrap/EmailContext.php
```diff
@@ -877,6 +878,27 @@ protected function assertEmailBcc($address)
         Assert::contains(current($this->matchingFakeEmails)['bcc_address'], $address, 'address not on bcc line');
     }
 
+    protected function assertEmailCc($address)
+    {
+        Assert::greaterThan(count($this->matchingFakeEmails), 0);
+        Assert::keyExists(current($this->matchingFakeEmails), 'cc_address');
```
briskt: I had to research `current` to refresh my memory. It seems to be deprecated in PHP 8.1 (if AI is correct). I also looked through this code and didn't find any use of `next` or `prev` so you should be able to simply replace `current($this->matchingFakeEmails)` with `$this->matchingFakeEmails[0]` since this array has numeric keys.

app/features/bootstrap/YiiContext.php
```diff
@@ -40,6 +40,10 @@ public function setupLogger()
 
         Yii::$app->set('emailer', $this->fakeEmailer);
 
+        // Reset between scenarios so these can't leak via the shared application instance.
+        Yii::$app->params['userMailAdminsCcOnInvite'] = false;
```
briskt: I'm very glad that you included these parameter resets! It might be slightly better to put them in an AfterScenario hook. However, it's not strictly necessary since every test that could use them would be affected by this hook. The difference is mostly semantic because this method name is `setupLogger` and not `resetParameters`. (Looks like feature creep had already set in before you got here!)

compose.yaml
```diff
@@ -159,6 +159,12 @@ services:
     working_dir: /dynamo
     command: bash -c "pwd && ls -al && php restart.php"
 
+  apimock:
```
briskt: Unless one Mockoon instance can be easily configured for multiple APIs, consider changing the name to `mail-admin` so other APIs (e.g. HIBP) could be mocked like this in the future.

compose.yaml
```diff
@@ -168,6 +174,8 @@ services:
     mem_limit: 200m
     working_dir: /app
     depends_on:
+      apimock:
+        condition: service_started
```
briskt: I don't know how quickly Mockoon starts up, but you could use a `healthcheck` to make it more robust. See an example here: https://github.com/mockoon/mockoon/tree/main/packages/cli#docker-compose

briskt (commented):
This is a good implementation. There are some changes I'd like to suggest.

app/common/components/MailAdmin.php
mtompset: I suspect if you do a `git blame`, you'll find @forevermatt did the split. Every method should do just one thing. If it was done by him, I'd leave it. Frequently, the split helps debug in the future.

briskt: `ApiClient` appears to be copied from ssp-base, but `MailAdmin` is new here. In `ssp-base` the equivalent code was added to the existing controller.

app/common/config/main.php
mtompset: Each API call to MailAdmin (the system that is my baby) has a separate API key. I believe ACCOUNT_MAIL_ADMINS_API_KEY would be better, since the call is to that API within MailAdmin. *shrug*

app/common/models/User.php
mtompset: Nested if's are a pain to debug.

samuel-sowpati-sil: Agreed. I'll refactor this block!

app/features/bootstrap/EmailContext.php
mtompset: -- @briskt the first index isn't necessarily 0.
[ 'a' => 'first entry', 'second entry', 'b' => 'third entry']
The first index is 'a', and the second one... I'm guessing 0 or 1, but it is some numeric.

https://www.php.net/manual/en/function.current.php -- calling on objects is deprecated.

Current is really ugly. How do we know there wasn't a next() somewhere in a long piece of code or as a side-effect of some function call?

```
@arrayKeys = array_keys($this->matchingFakeEmails);
$firstKey = $arrayKeys[0];
Assert::keyExists($this->matchingFakeEmails[$firstKey], 'cc_address');
Assert::contains($this->matchingFakeEmails[$firstKey]['cc_address'], $address, 'address not on cc line');
```
This is clearer to all code readers.

app/features/bootstrap/EmailContext.php
```diff
@@ -877,6 +878,27 @@ protected function assertEmailBcc($address)
...
+        Assert::contains(current($this->matchingFakeEmails)['cc_address'], $address, 'address not on cc line');
+    }
+
+    protected function assertEmailNotCc($address)
+    {
+        Assert::greaterThan(count($this->matchingFakeEmails), 0);
+        $ccAddress = current($this->matchingFakeEmails)['cc_address'] ?? '';
```
mtompset: Perhaps extracting a function to getCcAddress(): string might be useful?
Then the above function doesn't really need the exists and contains assertions.

app/common/components/MailAdmin.php
samuel-sowpati-sil: I've kept them separate —as ApiClient is just transport (same class as in ssp-base) and MailAdmin holds the domain logic, which keeps it out of the already-large User model. Glad to merge them if you feel strongly.

app/common/components/MailAdmin.php
samuel-sowpati-sil: Good call, the config now uses the ACCOUNT_MAIL_ADMINS_* prefix throughout

app/common/config/main.php
samuel-sowpati-sil: I went with ACCOUNT_MAIL_ADMINS_* to match the endpoint and keep each key distinct. Thanks!

app/common/components/SesMessage.php
samuel-sowpati-sil: Reverted `SesMessage` and moved the split into `Email::getMessage()`. That surfaced a latent bug — the Email model rejected a multi-address cc_address, so a real CC would've failed to queue — so I added a comma-aware validator plus a covering test.

app/features/bootstrap/EmailContext.php
samuel-sowpati-sil: I swapped current() for a small getCcAddress() helper that takes the first match via array_values(...)[0] — safe even though array_filter keeps the original keys. Thanks for the catch.

app/features/bootstrap/EmailContext.php
samuel-sowpati-sil: Mark, Ive added getCcAddress(); the three CC assertions now just check what it returns.

app/features/bootstrap/YiiContext.php
samuel-sowpati-sil: Thanks Steve. I pulled the resets into a dedicated #[BeforeScenario] resetParameters(). Kept it a `Before` hook so each scenario starts from a known state regardless of run order.

compose.yaml
samuel-sowpati-sil: I've renamed the service to mail-admin

app/common/models/Email.php
```diff
@@ -170,7 +197,7 @@ public function getMessage()
          */
 
         if ($this->cc_address) {
-            $mailer->setCc($this->cc_address);
+            $mailer->setCc(array_map('trim', explode(',', $this->cc_address)));
```
mtompset: I know it is more vertical space, but this many nested functions is a PR-review parsing nightmare. I would break them out. The code is okay, but yucky.

app/features/bootstrap/EmailClientContext.php
```diff
@@ -251,6 +251,24 @@ public function testGetMessageRendersAsHtmlAndText()
...
+
+        $email->attributes = [
+            'to_address' => $this->stubToAddress,
+            'cc_address' => 'cc1@example.org,cc2@example.org',
+            'subject' => $this->stubSubject,
+            'text_body' => $this->stubTextBody,
+            'html_body' => $this->stubHtmlBody,
+        ];
+
+        Assert::true($email->save(), current($email->getFirstErrors()));
```
mtompset, line 267: Ewwww. Surely there is a better way to give meaningful error feedback.

mtompset (approved):
I think this is okay, but it's been a long day. :)

app/common/models/Email.php
samuel-sowpati-sil: Good call Mark. I pulled the split into a small `splitAddresses()` helper, so `setCc()` (and the validator) just call that instead of nesting array_map/explode inline.

app/features/bootstrap/EmailClientContext.php
samuel-sowpati-sil: Fair point. But `current($email->getFirstErrors())` is the existing pattern used in ~9 other spots in this file, so I matched it here for consistency. Happy to switch them all to something clearer like `implode('; ', $email->getFirstErrors())` in a follow-up PR

mtompset: Okay, since this isn't a completely new file.

app/common/components/MailAdmin.php
briskt: No, it's fine. I prefer simplicity over abstractions. But it's just an opinion.

app/features/bootstrap/YiiContext.php
```diff
@@ -43,6 +43,14 @@ public function setupLogger()
         $this->addFakeLogTarget();
     }
 
+    #[BeforeScenario]
+    public function resetParameters()
+    {
+        // Reset between scenarios so these can't leak via the shared application instance.
+        Yii::$app->params['accountMailAdminsCcOnInvite'] = false;
```
briskt: Using a BeforeScenario renders the default values in test.env ineffective. Consider removing them from test.env or changing this to save the value in BeforeScenario and restore it in AfterScenario rather than hard-code it here. Just thinking it should have one source of truth, avoiding the confusing situation where a person tries to change it in test.env and not getting the desired effect.

briskt (approved):

app/features/bootstrap/YiiContext.php
samuel-sowpati-sil: Thanks Steve — good principle. I went with your 2nd suggestion. The config defaults in `common/config/main.php` (and anything in `test.env`) are now the single source of truth — the hooks just save the pre-scenario value and put it back, keeping scenarios isolated without duplicating `false`/`''`.

briskt (approved):

PR #480, opened 2026-06-17
Update expires_on database column rather than "on display"

jason-jackson:
[IDP-128](https://support.gtis.sil.org/issue/IDP-128) Update expires_on database column rather than "on display"

---

### Changed
- Update expires_on database column rather than "on display"
- PSR2 fixes

---

### PR Checklist
- [ ] Documentation (README, local.env.dist, openapi.yaml, etc.)
- [x] Run `make composershow`
- [x] Run `make psr2`

### DIFF
--- a/app/common/models/Mfa.php
+++ b/app/common/models/Mfa.php
@@ -149,6 +149,11 @@ class Mfa extends MfaBase
                 $this->user->scenario = User::SCENARIO_UPDATE_USER;
                 $this->user->save();
             }
+
+            // Update password expires_on and grace_period_ends_on if any verified mfa options
+            if ($this->user->getVerifiedMfaOptionsCount() == 1 && $this->user->currentPassword !== null) {
+                $this->user->currentPassword->save();
+            }
         }
     }
 
@@ -196,7 +201,10 @@ class Mfa extends MfaBase
             $this
         );
 
-        $this->user->extendGracePeriodIfNeeded();
+        // Check if any MFA, and roll back expiration date and update grace period as needed
+        if ($this->user->getVerifiedMfaOptionsCount() == 0 && $this->user->currentPassword !== null) {
+            $this->user->currentPassword->save();
+        }
     }
 
     /**
--- a/app/common/models/Mfa.php
+++ b/app/common/models/Mfa.php
@@ -152,7 +152,7 @@ class Mfa extends MfaBase
 
             // Update password expires_on and grace_period_ends_on if any verified mfa options
             if ($this->user->getVerifiedMfaOptionsCount() == 1 && $this->user->currentPassword !== null) {
-                $this->user->currentPassword->save();
+                $this->user->currentPassword->updateExpiry();
             }
         }
     }
@@ -203,7 +203,7 @@ class Mfa extends MfaBase
 
         // Check if any MFA, and roll back expiration date and update grace period as needed
         if ($this->user->getVerifiedMfaOptionsCount() == 0 && $this->user->currentPassword !== null) {
-            $this->user->currentPassword->save();
+            $this->user->currentPassword->updateExpiry();
         }
     }
 
--- a/app/common/models/Password.php
+++ b/app/common/models/Password.php
@@ -199,14 +199,14 @@ class Password extends PasswordBase
 
     /**
      * Update expires_on and grace_period_ends_on fields
-     * @return void
+     * @return bool
      */
     public function updateExpiry()
     {
         $this->expires_on = self::expires()();
         $this->grace_period_ends_on = self::gracePeriodEnds()();
         $this->setScenario(self::SCENARIO_UPDATE_METADATA);
-        $this->save();
+        return $this->save();
     }
 
     /**
--- a/app/common/models/Password.php
+++ b/app/common/models/Password.php
@@ -123,8 +123,8 @@ class Password extends PasswordBase
     private function expires(): Closure
     {
         return function () {
-            if (!empty($this->expires_on)) {
-                return $this->expires_on;
+            if ($this->hibp_is_pwned === 'yes') {
+                return MySqlDateTime::relativeTime('+5 minutes');
             }
 
             $lifespan = Yii::$app->params['passwordLifespan'];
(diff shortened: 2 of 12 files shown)


app/common/models/Mfa.php
```diff
@@ -149,6 +149,11 @@ public function afterSave($insert, $changedAttributes)
                 $this->user->scenario = User::SCENARIO_UPDATE_USER;
                 $this->user->save();
             }
+
+            // Update password expires_on and grace_period_ends_on if any verified mfa options
+            if ($this->user->getVerifiedMfaOptionsCount() == 1 && $this->user->currentPassword !== null) {
```
briskt, line 154: Why compare against `1` and not `>= 1`?

app/common/models/Password.php
```diff
@@ -121,18 +123,51 @@ public function behaviors(): array
     private function expires(): Closure
     {
         return function () {
+            $oldHIPB = $this->getOldAttribute('hibp_is_pwned');
+            if ($this->hibp_is_pwned === 'yes' && $oldHIPB !== 'yes') {
+                return $this->expires_on;
+            } elseif ($this->hibp_is_pwned === 'yes') {
+                return MySqlDateTime::relativeTime('+5 minutes');
```
briskt, line 131: Why 5 minutes?

app/common/models/Password.php
```diff
@@ -121,18 +123,51 @@ public function behaviors(): array
     private function expires(): Closure
     {
         return function () {
+            $oldHIPB = $this->getOldAttribute('hibp_is_pwned');
+            if ($this->hibp_is_pwned === 'yes' && $oldHIPB !== 'yes') {
```
briskt, line 127: Why does the old value affect this?

app/common/models/Password.php
```diff
@@ -163,38 +198,33 @@ public function attributeLabels()
...
-        $dateIso = $date . 'T23:59:59Z';
-        if ($this->user->getVerifiedMfaOptionsCount() > 0) {
-            $extended = strtotime(\Yii::$app->params['passwordMfaLifespanExtension'], strtotime($dateIso));
-            return Utils::getIso8601($extended);
-        }
-        return $dateIso;
+        $this->expires_on = self::expires()();
+        $this->grace_period_ends_on = self::gracePeriodEnds()();
+        $this->setScenario(self::SCENARIO_UPDATE_METADATA);
+        $this->save();
```
briskt: Probably should check the return value to see if it succeeded.

app/console/migrations/m260527_000000_update_password_expiration_grace_period.php
```diff
@@ -0,0 +1,91 @@
...
+        $this->execute($sql);
+    }
+
+    /**
+     * Convert a simple PHP relative string like "+1 year" or "-2 months" to a MySQL interval fragment:
+     * returns e.g. "+ INTERVAL 1 YEAR" or "- INTERVAL 2 MONTH", or null if it can't parse.
+     *
+     * Supports: seconds/minutes/hours/days/weeks/months/years (singular/plural/short forms).
+     */
+    protected function mysqlInterval(string $s): ?string
```
briskt: This function and invertIntervalSql maybe should be in helpers/MySqlDateTime. And have unit tests to give some degree of assurance that they work correctly.

app/features/bootstrap/SesMailerUnitTestsContext.php
```diff
@@ -23,7 +23,7 @@ class SesMailerUnitTestsContext extends YiiContext
     #[Given('I have a new SesMailer')]
     public function iHaveANewSesMailer()
     {
-        $this->mailer = new class extends SesMailer {
+        $this->mailer = new class () extends SesMailer {
```
briskt, line 26: This is weird.

briskt (approved):

app/common/models/Password.php
briskt: At least with Go it's easy for lint to warn you about ignoring an error.

app/common/models/Mfa.php
jason-jackson: Two reasons:
- L137 really only ever means it applies to 1
- Anything after the first we don't extend expires_on any further, so it's not needed.

app/common/models/Password.php
jason-jackson: Old value means that it is already marked as pwned and we should not update the expires_on.

app/features/bootstrap/SesMailerUnitTestsContext.php
jason-jackson: I concur, but it's what `make psr2` did.

app/common/models/Password.php
jason-jackson: Taken from L276/306 in this file.

app/common/models/Password.php
jason-jackson: Adding this, but I think if save fails it throws an exception (I could be wrong).

app/common/models/Mfa.php
briskt: That's fine if you're sure it will always and only trigger this code as expected. My concern was that subtle changes or unexpected behavior could cause it to skip the update.

app/common/models/Password.php
briskt: Isn't the update method idempotent -- it can be called many times without problems? If not, it probably shouldn't be in the before update hook.

app/common/models/Password.php
briskt: It may throw exceptions for some failures but I am certain it returns false for validation failure.

app/common/models/Password.php
briskt: Ah, thanks. Since it's in two places (and because it is mysterious) it might help to make it a constant.

app/common/models/Mfa.php
jason-jackson: That's a valid concern, but I can't think of any cause for failures right now, granted, it wouldn't be unexpected if I could.

app/common/models/Password.php
jason-jackson: This should be able to be called many times without problem still. I could add a check to updateExpiry though to not bother if it is pwned.

jason-jackson: Hmm, thought more about it, since I'm calling the method directly on updateExpiry, I can remove the following lines (L108, L116)

```
ActiveRecord::EVENT_BEFORE_UPDATE => 'expires_on',
...
ActiveRecord::EVENT_BEFORE_UPDATE => 'grace_period_ends_on',
```

And then I no longer need the hibp check in expires() as it no longer cares as it won't have an old value on creation.

jason-jackson: Tests failed, so I'm just leaving it as is.

briskt (approved):

PR #488, opened 2026-07-02
Update grace_period_ends_on to fix login

jason-jackson:
[IDP-128](https://support.gtis.sil.org/issue/IDP-128) Update expires_on database column rather than "on display"

---

### Fixed
- Migrate grace_period_ends_on to fix login

---

### PR Checklist
- [ ] Documentation (README, local.env.dist, openapi.yaml, etc.)
- [ ] Run `make composershow`
- [ ] Run `make psr2`

### DIFF
--- a/app/console/migrations/m260702_000000_update_password_grace_period.php
+++ b/app/console/migrations/m260702_000000_update_password_grace_period.php
@@ -1,0 +1,43 @@
+<?php
+
+use common\helpers\MySqlDateTime;
+use yii\db\Migration;
+use yii\db\Expression;
+
+class m260702_000000_update_password_grace_period extends Migration
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function safeUp()
+    {
+        $interval = MySqlDateTime::interval(\Yii::$app->params['passwordMfaLifespanExtension']);
+
+        $sql = "
+            UPDATE password
+            SET grace_period_ends_on = grace_period_ends_on {$interval}
+            WHERE
+                password.hibp_is_pwned = 'no'
+                AND EXISTS (SELECT 1 FROM mfa WHERE mfa.user_id = password.user_id AND mfa.verified = 1)
+        ";
+        $this->execute($sql);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function safeDown()
+    {
+        $interval = MySqlDateTime::interval(\Yii::$app->params['passwordMfaLifespanExtension']);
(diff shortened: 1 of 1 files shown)


app/console/migrations/m260702_000000_update_password_grace_period.php
```diff
@@ -0,0 +1,43 @@
...
+use yii\db\Expression;
+
+class m260702_000000_update_password_grace_period extends Migration
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function safeUp()
+    {
+        $interval = MySqlDateTime::interval(\Yii::$app->params['passwordMfaLifespanExtension']);
```
briskt: Would it be better to set it relative to the password expiration?

jason-jackson: Good question, I'm not sure. I think it's approximately six of one, half a dozen of the other. This way keeps it inline with expires_on migration, so theoretically they should be the same.

briskt: My concern is potentially bumping the expires_on twice, but I'm not sure if that would actually happen. Setting it relative to expiry would be idempotent.

briskt: Not that this migration would run twice, but if some users' grace period is already correct for some reason (e.g. they reset their password between the last migration and this one).

jason-jackson: This look fine to you?

```
        $interval = MySqlDateTime::interval(\Yii::$app->params['passwordGracePeriodExtension']);

        $sql = "
            UPDATE password
            SET grace_period_ends_on = expires_on {$interval}
```

briskt: Yep

app/console/migrations/m260702_000000_update_password_grace_period.php
```diff
@@ -0,0 +1,43 @@
...
+        $this->execute($sql);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function safeDown()
+    {
+        $interval = MySqlDateTime::interval(\Yii::$app->params['passwordGracePeriodExtension']);
+        $interval = MySqlDateTime::invertInterval($interval);
```
briskt: Shouldn't this still be positive?

briskt (approved):

PR #493, opened 2026-07-17
Cleanup

briskt:
### Fixed
- style cleanup
- update dependencies.json for recent patches

### DIFF
--- a/app/common/exceptions/InvalidCodeException.php
+++ b/app/common/exceptions/InvalidCodeException.php
@@ -2,6 +2,4 @@
 
 namespace common\exceptions;
 
-class InvalidCodeException extends \Exception
-{
-}
+class InvalidCodeException extends \Exception {}
--- a/app/console/controllers/GaController.php
+++ b/app/console/controllers/GaController.php
@@ -4,6 +4,4 @@ namespace console\controllers;
 
 use yii\console\Controller;
 
-class GaController extends Controller
-{
-}
+class GaController extends Controller {}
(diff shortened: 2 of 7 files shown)


app/common/exceptions/InvalidCodeException.php
```diff
@@ -2,6 +2,4 @@
 
 namespace common\exceptions;
 
-class InvalidCodeException extends \Exception
-{
-}
+class InvalidCodeException extends \Exception {}
```
mtompset, line 5: This is PER compliant, but not PSR-12 compliant.

app/console/controllers/GaController.php
```diff
@@ -4,6 +4,4 @@
 
 use yii\console\Controller;
 
-class GaController extends Controller
-{
-}
+class GaController extends Controller {}
```
mtompset, line 7: Are we migrating projects to PER standard?

mtompset (approved):
Just want to make sure you are aware of PSR-12 vs. PER style differences. :)

app/common/exceptions/InvalidCodeException.php
briskt: The change comes from php-cs-fixer, which controls the formatting rules for this project.

app/console/controllers/GaController.php
briskt: I don't know. I just go with the flow as best as I can.

PR #499, opened 2026-07-24
Confirm people cannot change their password to an old password

samuel-sowpati-sil:
[IDP-1912](https://support.gtis.sil.org/issue/IDP-1912) Confirm people cannot change their password to an old password

### Added
- reuse coverage: reject change to current password, reject change to a previous password, 409 on assessing a reused password, and that a limit of 0 disables the check.

### Changed (non-breaking)
- `local.env.dist`: note that `PASSWORD_REUSE_LIMIT=0` disables the reuse check.

### DIFF
--- a/app/common/models/Password.php
+++ b/app/common/models/Password.php
@@ -76,6 +76,11 @@ class Password extends PasswordBase
     {
         $reuseLimit = Yii::$app->params['passwordReuseLimit'];
 
+        // A numeric 0 becomes LIMIT 0, which checks no rows and silently allows reuse of the current password.
+        if (is_numeric($reuseLimit) && (int) $reuseLimit === 0) {
+            $reuseLimit = 1;
+        }
+
         /** @var Password[] $passwords */
         $passwords = Password::find()->where(['user_id' => $this->user_id])
                                      ->orderBy(['id' => SORT_DESC])
--- a/app/common/models/Password.php
+++ b/app/common/models/Password.php
@@ -76,11 +76,6 @@ class Password extends PasswordBase
     {
         $reuseLimit = Yii::$app->params['passwordReuseLimit'];
 
-        // A numeric 0 becomes LIMIT 0, which checks no rows and silently allows reuse of the current password.
-        if (is_numeric($reuseLimit) && (int) $reuseLimit === 0) {
-            $reuseLimit = 1;
-        }
-
         /** @var Password[] $passwords */
         $passwords = Password::find()->where(['user_id' => $this->user_id])
                                      ->orderBy(['id' => SORT_DESC])
--- a/app/features/bootstrap/UnitTestsContext.php
+++ b/app/features/bootstrap/UnitTestsContext.php
@@ -35,8 +35,11 @@ class UnitTestsContext extends YiiContext
     /** @var User */
     protected $tempUser;
 
-    /** @var bool whether the most recent password submission was saved */
-    protected $passwordWasAccepted;
+    /** @var string */
+    protected $originalPassword;
+
+    /** @var ConflictHttpException|null */
+    protected $reuseException;
 
     /** @var bool whether the Mfa option is considered newly verified */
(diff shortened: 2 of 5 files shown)


app/common/models/Password.php
```diff
@@ -76,6 +76,11 @@ private function hasAlreadyBeenUsedTooRecently(): bool
     {
         $reuseLimit = Yii::$app->params['passwordReuseLimit'];
 
+        // A numeric 0 becomes LIMIT 0, which checks no rows and silently allows reuse of the current password.
```
briskt: As far as I recall, we never intentionally use a setting of 0 but it seems like that would be the value to use if you want to disable this check. Isn't that the current behavior?

samuel-sowpati-sil: You're right. Setting the limit to 0 turns the check off completely, and my change makes 0 act like 1 (still blocks reusing current password).

My only worry was that 0 isn't documented anywhere and looks like it should mean "strictest," not "off". So someone could disable the check by accident.

Which approach do you think would be best?:

1.Keep 0 as the off-switch, but document it and change the test to expect that.
2. Keep my change, so 0 still blocks the current password.

briskt: Since we have no other way to disable the feature I think it's best to leave it as-is. I don't necessarily think it must be documented but you're welcome to do so if you like.

app/features/password.feature
```diff
@@ -14,6 +14,60 @@ Feature: Password
...
+      And the requester is authorized
+      And I add a user with an employee_id of "10000"
+      And I provide the following valid data:
+        | property | value              |
+        | password | k23@U$%235u25@I2$o |
+      And I request "/user/10000/password" be updated
+      And the response status code should be 200
+    When I request "/user/10000/password" be updated
+    Then the response status code should be 409
+      And the response body should contain "May not be reused yet"
```
forevermatt: Rather than long, API-language wording here, let's keep it as clear as possible. You can see the previous two Scenarios as examples.

This could be changed to something like...
```suggestion
  Scenario: Attempt to change a password to the one currently in use
    Given there is a user in the database
      And the user has a password
      And the password reuse limit is 10
    When the user tries to change their password to that same password
    Then a 409 error of "May not be reused yet" should be returned
```

The wording of the next two Scenarios could be similarly simplified. (It sounds like the "reuse limit of zero" Scenario may need to be changed or removed.)

You may already know this, but it's also helpful to run `behat` with the `--append-snippets` option so that it generates the stub functions and all you have to write by hand is the Gherkin and the implementation details within those stub functions.

app/features/bootstrap/UnitTestsContext.php
forevermatt, line 1: I like how clean your test-implementation details are here. 👍

forevermatt (commented):
Thanks for working on this

app/features/bootstrap/UnitTestsContext.php
```diff
@@ -442,6 +449,76 @@ public function aNewPasswordHashShouldBeStored()
...
+
+        try {
+            $this->tempUser->assessPassword($this->originalPassword);
+        } catch (ConflictHttpException $e) {
+            $this->reuseException = $e;
+        }
+    }
+
+    #[Then('a 409 error of :message should be returned')]
+    public function a409ErrorOfShouldBeReturned($message)
```
forevermatt: Interesting... I expected Behat to also treat the number as another parameter to this test step. It's fine as-is, I just expected it to be less hard-coded.

forevermatt (approved):

app/features/bootstrap/UnitTestsContext.php
samuel-sowpati-sil: Initially I wrote it manually. Now, I ran --append-snippets and included its parameterized form ( `:statusCode` ) in the following commit. The `--append-snippets` command is certainly helpful :smile:

briskt (approved):

PR #510, opened 2026-08-20
replace phpMyAdmin with Adminer

briskt:
[IDP-2128](https://support.gtis.sil.org/issue/IDP-2128) Switch IDP over to use Adminer

---

### Changed (non-breaking)
- Replace phpMyAdmin with Adminer in the local docker compose environment.


compose.yaml
```diff
@@ -196,35 +196,20 @@ services:
...
+  adminer:
+    image: adminer
     ports:
-      - "51142:80"
-    depends_on:
-      - testdb
-    env_file:
-      - path: ./local.env
-        required: false
+      - "51141:8080"
```
hobbitronics: should this be 51142? The port is also different so I supsect it's deliberate, but that will cause collisions if someone is running a local environment while runnign tests right?

hobbitronics (approved):

compose.yaml
briskt: It's 51141 because that was the port for 'phpmyadmin' (see old line 199). The internal port is 8080 because that's what the adminer image uses. No, it won't conflict because it's only used inside this particular container.

